PR #16, opened 2023-05-26
1 create init function

### LINKED ISSUE #1: Create Init Function
When starting the shell, it should initialize its environment. 

- This will probably be implemented with the environment pointer char **envp, so that the main function looks like main(int argc, char **argv, char **envp)
- Alternative is then environ variable - which would be our first global

The following situations should be handled:

1. There was an environment --> copy it
2. There was no environment --> what then? (what is when home is not set?)

gwolf-011235:
0. Done too much for this one branch - now in the right direction :D

1. Implemented an environment import. 

- The environment is imported from the global variable "environ" which is extern defined in minishell.h. This seems to be more portable than main(int argc, char **argv, char **envp). 
- A hashtable env_table is setup to save the strings. The table receives a size and hash function as well as an elements pointer.
- The table uses external chaining: if an index is already filled the same index is used, via a linked list.
- Every env_var (which is a string) from environ is ft_strdup into the hashtable.
- The hash is generated from the variable name (e.g. PWD) which is received via strchr(=) 
- Possible BUG: if the string has no "=" program dies
- If no environment is received or PWD and/or SHLVL is not set, they are created.
- PWD via getpwd
- SHLVL is set to 1 if not found
- Env_var _ (Underscore) is missing right now
- A simple print function can print out the variable, there also exists an lookup function to search and a delete function to delete one entry
- The table can also be destroyed, there seem to be no leaks right now.

2. Setup simple error handling

- A struct enum t_error is implemented to standardize all error codes.
- The idea is that most functions use t_error as return value. They return SUCCESS if it worked or one of the defined error codes. They can be easily checked and printed.

3. Setup Doxygen documentation

- The subdirectory docs/ is created for documentation
- Right now there is a Doxygen config file which generates the docs
- All functions and struct in this branch are commented to generate doc
- doxygen-awesome-css is implemented to make the docs prettier
- PROBLEM: there doesn't seem to be a way to comment the fields of structs outside of the struct itself. This is not compatible with norminette. Right now workaround is to comment them as @param. Except for the enum t_error: There the comments are in the wrong scope.


u413-284-si:
looks good, performed rebase with main

PR #19, opened 2023-06-06
3 terminal prompt

### LINKED ISSUE #3: Terminal Prompt
The shell displays a terminal prompt like 

_gwolf@host:/home/gwolf$_

It needs the following

- [ ] Username and Machine
- [ ] Current working directory
- [ ] More customizations?
- [ ] Tilde (~) instead of $HOME directory

gwolf-011235:
Closes #3 

1. Implementation of terminal prompt

Terminal prompt gets created using $PS1 (for the standard one) and $PS2 (for the input one, e.g. while heredoc). Search for  env_var. If found, str_dup'd. Then try to expand. Right now the following escape sequences are supported:
- \u: user
- \h: hostname
- \w: current directory where $HOME gets replaced with tilde (~)
- \n: newline
- \r: carriage return
For expansion look for a known escape sequence in the string (the "token). If found try to get a corresponding replacement string. Then calc len of expanded string with replacement, malloc it and copy the right bytes. 

2. Readline

Readline is called in main but doesn't do much. Gets prompt (right now hardcoded PS1) and wants input from user. Then it repeats it back. Just needed to link libreadline in Makefile and it just works.

3. Setting up testing

A new subdir test. The idea is that the tests are functions and the test_main is just a driver to call them. Simple implementation but easy to use. "make test" compiles sources plus the testfiles into binary "tester". To avoid two mains the OG main has a macro, which gets activated by "make test" (this needs to be removed before norminette). Before running "make test" clean needs to be run to remake the OG main. 
 
4. Addition of hashtable_swap

Small function to swap the string of an existing env_var. Faster than delete + insert.


u413-284-si:
Nice prompt, although we need to verify its functionality on different system setups (virtual machines, direct installs, different OS) as behaviour seems to vary. Furthermore some issues with test suite  -> handling needs to be streamlined (libft compiled directly with 'make test'). But nice idea which will serve well in the following development.

PR #21, opened 2023-06-24
20 error handling while lexing

### LINKED ISSUE #20: Error handling while lexing
Errors to handle

- [x] Quotes - double and single
- [x] Empty input
- [x] Wrong pipe
- [x] Wring redirection

gwolf-011235:
Closes #20 

1. Implements function ft_check_syntax() which checks the input string for

- Quotes: If a quote (single or double) is found check the input for another one. If found jump over the quoted part. If a single quote is quoted with double quotes it would get jumped over and vice versa
- Pipes: Check if a single pipe is at the start or at the end -> error. Also if double pipe (only needed for bonus)
- Redirect: Check after a redirect symbol < or >. Allowed is another one of same kind (<< or >>); not allowed are pipe, other kind, zero terminator --> could be reduced further - e.g. semicolon as filename would be silly 

2. Tests for this function in test dir


u413-284-si:
End of input should be \n instead of \0, no?

gwolf-011235:
Thought so too but readline cuts of the final newline. 
From man page: "The line returned has the final newline removed, so only the text of the line remains."

u413-284-si:
If the final \n is cut and only the text remains, that doesn't mean that there will be a NULL character at the end. It is just the last input character. This is how I interpret it.

u413-284-si:
Due to merge of branch 18 conflicts need to be resolved before merge is possible.

gwolf-011235:
rebased onto main, no conflicts left.

NB: the rename of test to tests lead to a conflict - maybe we should not rename stuff while having different branches :D

u413-284-si:
Such a lovely conflict. Interesting though, shouldn't git be intelligent enough to understand that difference?!

gwolf-011235:
To be more precise: it was the file test_check_syntax.c which was created in test/
Since it was renamed to tests/ git didn't know where to put it. I think the VSCode git integration might also not be optimal - maybe terminal would have been easier :D

u413-284-si:
mamma mia

PR #23, opened 2023-06-25
18 makefile is sometimes strange

### LINKED ISSUE #18: Makefile is sometimes strange
Today Makefile was recompiling every time. I think it has something to do with dependency generation. Couldn't find out what was wrong; now it works again

gwolf-011235:
### 1. Improve Makefile

Improvements and eye candy. This should close #18 

#### 1.1. Better dependencies:

Dependency generation is now protected against old clang shenanigans. The .d files are created as .Td (for temporary). With $(POSTCOMPILE) in the compile step they are moved to .d files and then the compiled object gets touched, ensuring the right timestamp. See link in wiki

#### 1.2. Target test is now easier to use:

It removes the "normal" main to recompile with TESTING (effectivly renaming main). Now the test_main can be compiled and linked, making it the only main. After compile and link the renamed main gets again deleted, making it possible to use "make" and "make test" right after another.

#### 1.3. Target debug renamed to leak:

Old debug target added -g flag - this is now in normal compiling. New leak target adds fsanitize=leak to compile and link. It has to clean to recompile everything. With this flag the leak checker of google gets integrated.

### 2. Rename subdirectory test to tests

Since the target test and the directory have the same name some things didn't work out. Rename was easiest fix.


u413-284-si:
Wildcards are not allowed in Makefile according to norminette. 

Should have a call to discuss in more detail the changes in testing and debug. Tomorrow? What time would be good?

gwolf-011235:
Norm doesn't mention wildcards afaik. 

Last rule in part about Makefile mentions that every source file has to be listed (effectively means no wildcard for source files).

But wildcard for dependency files should be no problem. 
Is a protection, so can also be removed if too spicy.

u413-284-si:
True,
it doesn't explicitly forbid wildcards. Therefore we can leave it and see if it does its job correctly. In the beginning I had a wildcard in my Makefile (2nd rush) and it exhibited weird behaviour. Shall observe closely.

u413-284-si:
Acknowledged after review

PR #25, opened 2023-07-07
24 prep envp for execve

### LINKED ISSUE #24: Prep envp for execve
In order to be able to pass an envp to execve the current env_vars need to be extracted from the hashtable and put into a char **array.

This is also an opportunity to rename the env_tab into sym_tab as foreshadowed (terminology: the environment is saved in a symbol table) and update doxygen comments.

gwolf-011235:
Simple addition of two functions which create and destroy a char **array (envp). This can be passed to execve.

1. New var num_elements in t_hashtable
Hashtable now has a var to save the number of elements. The functions ft_hashtable_insert() and ft_hashtable_delete() increase and decrease this counter.

2. ft_envp_create()
Malloc an array of size num_elements plus 1 (for NULL termination). Copy the pointers from hashtable into this array. Last element is set to NULL

3. ft_envp_destroy()
Check if its exists. If yes free (dont touch the elements themselves), and set to NULL. 

4. Idea: The envp only needs to be destroyed, if environment variables are added / deleted (i.e. it gets bigger/smaller). Maybe we can implement a check before creation of envp, if the environment has changed. If only the values have changed, then only the pointers need to be updated.

Closes #24

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
 			"name": "test",
 			"type": "cppdbg",
 			"request": "launch",
-			"program": "${workspaceFolder}/tester",
+			"program": "${workspaceFolder}/test",
 			"args": [],
 			"stopAtEntry": true,
 			"cwd": "${fileDirname}",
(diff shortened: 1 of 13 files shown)


u413-284-si:
approved, whether we check before destroying can be decided later on, if we see a benefit in this extra work

PR #26, opened 2023-07-12
5 lexer

### LINKED ISSUE #5: Lexer
Part of input processing

> a software component that takes a string and breaks it down into smaller units that are understandable by a language. These smaller units are called lexical tokens or lexemes. In other words, you can think of a lexer as a black box that takes a sentence as input and breaks it into smaller units —essentially, words. [LINK](https://medium.com/young-coder/how-i-wrote-a-lexer-39f4f79d2980)

It would be nice to store every token it finds in a struct. Creating a linked list of tokens which can then be used by the parser.

:dancing_men:

u413-284-si:
Provides lexing of input string deliverd by readline.

1. The input string is saved in a src struct containing additionally its size and the current pointer position. This enables the lexer to move character by character through the input string and remember its last position interacted with.
2. Tokenising: the characters are consecutively saved in a temporary buffer string. The saving process is halted when spaces, tabs, newlines or pipes are encountered. Then a token is created and the buffer string saved into it. 
3. The token is saved into a newly created node which is being added to a token list. The list is implemented as a double-linked list in order to provide the possibility to go forth and backwards in regards to the following parser. Then the tokeniser is initiated again to retrieve the next token out of the src string.
4. In case of newlines and pipes, the tokeniser does not skip them, but saves them as individual tokens in the list.
5. Testing: Different test cases have been produced

- cmd without newline
- cmd with newline
- cmd with option
- cmd with several spaces
- cmd with in and outfiles
- several cmds with in and outfiles piped
- piping without spaces between cmds

7. lexer.c & lexer.h: prototypes for following parser executing the categorisation of the individual tokens

closes #5


src/lexer_tok.c
```diff
@@ -0,0 +1,136 @@
...
+ */
+t_error	ft_create_tok(t_tok *token, char *s)
+{
+	if (!token || !s)
+		return (ERR_EMPTY);
+	token->tok_size = ft_strlen(s);
+	token->tok = ft_realloc (token->tok, token->tok_size + 1, 1);
+	if (!token->tok)
+	{
+		free(token->tok);
```
gwolf-011235: This would segfault since the if clause checks for NULL pointer in token->tok

src/lexer_tok.c
```diff
@@ -0,0 +1,136 @@
...
+	buf = malloc(buf_size);
+	if (!buf)
+		return (ERR_MALLOC);
+	buf_pos = 0;
+	buf[0] = '\0';
+	err = ft_partition(src, buf, &buf_pos, &buf_size);
+	if (buf_pos >= buf_size)
+		buf_pos--;
+	buf[buf_pos] = '\0';
+	if (err == ERR_EOF)
```
gwolf-011235, line 74: Why do we check if err == EOF when in both cases the same function ft_create_tok() with the same args gets called?

src/lexer_utils.c
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * @brief Delete the content of the node.
+ * 
+ * @param s Content of node.
+ */
+void	ft_del_content(void *s)
+{
+	if (!s)
+		return ;
+	s = "";
```
gwolf-011235: This just sets the address of the pointer s to an empty string literal. But since this is just a copy of the passed pointer (param of the function) it doesn't do anything :O 
Maybe just free()?
I think we can even get rid of the function pointer alltogether since we just want to free stuff

src/lexer_utils.c
```diff
@@ -0,0 +1,103 @@
...
+ * @brief Delete the content and the node itself.
+ * 
+ * @param lst List containing the nodes to be deleted.
+ * @param del Pointer to function to delete content of nodes.
+ */
+void	ft_del_node(t_tkn_list *lst, void (*del)(void*))
+{
+	if (!lst || !del)
+		return ;
+	(*del)(lst -> content);
```
gwolf-011235: I think we don't need the function pointer if it should just free the content string. If it shouldn't free it, even better :D

Makefile
```diff
@@ -98,7 +104,7 @@ $(TEST): prep_test $(TEST_OBJS) $(OBJS)
 	$(CC) $(LDFLAGS) $(TEST_OBJS) $(OBJS) $(LDLIBS) -o $@
 	rm -f obj/main.o
 	@printf "$(GREEN)Starting test!$(RESET)\n"
-	./$(TEST)
+#	./$(TEST)
```
gwolf-011235: Should be just get rid of the automatic running? So if it doesn't compile it doesn't get deleted?

Makefile
```diff
@@ -163,4 +169,14 @@ tclean:
 re: fclean all
 
 # Include the dependency files that exist. Use wildcard to avoid failing on non-existent files.
-include $(wildcard $(DEPFILES))
\ No newline at end of file
+include $(wildcard $(DEPFILES))
```
gwolf-011235, line 171: This line should always be the last line, so that the dependency generation works.

in
gwolf-011235, line 1: I think we can just delete this file :D

inc/lexer_tok.h
```diff
@@ -0,0 +1,35 @@
...
+/*   Updated: 2023/07/12 22:42:45 by sqiu             ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#ifndef LEXER_TOK_H
+# define LEXER_TOK_H
+
+/* ====== Includes ====== */
+
+//# include "minishell.h"
```
gwolf-011235: General: there are some lines with commented out minishell.h. We need to refactor minishell.h anyway but we shouldn't forget to delete these :D

inc/minishell.h
```diff
@@ -67,11 +127,17 @@ typedef struct s_checks {
...
+ * @param token Struct containing the current token and its size.
+ * @param lst_head Pointer to first node of token list.
+ * @param cmds List of structs containing simple commands.
  */
 typedef struct s_data {
 	t_hashtable	*env_table;
 	t_checks	checks;
 	t_error		err;
+	t_tok		token;
+	t_tkn_list	*lst_head;
```
gwolf-011235, line 139: I think we don't need t_tok and t_tkn_list in t_data. 
If we tokenize and then pass the token list to the ft_fill_cmd_struct (which doesn't exist yet) then we don't need to add it to data. I don't think we need to return to main in between the steps. If something (malloc error) happens between tokenize and fill cmd we can't do anything anyway

inc/test.h
```diff
@@ -18,6 +18,7 @@
 void	test_hashtable(void);
 void	test_replace_token(void);
 void	test_prompt(void);
+void	test_lexer(void);
```
gwolf-011235, line 21: I think test_tokenize would be more fitting.

out
gwolf-011235, line 1: Delete?

src/ft_memory.c
```diff
@@ -41,9 +41,9 @@ void	*ft_realloc(void *ptr, size_t new_size, size_t curr_size)
 	new_ptr = malloc(new_size);
 	if (new_ptr != NULL)
 	{
-		if (curr_size == 0)
+		if (curr_size != 0)
 			ft_memcpy(new_ptr, ptr, curr_size);
 		free(ptr);
 	}
 	return (new_ptr);
```
gwolf-011235, line 48: I think this could also be a t_error function. We would need to pass in a pointer pointer to void and the set it to the newly malloc'd address. It can return SUCCESS and ERR_MALLOC

src/lexer.c
gwolf-011235, line 1: This is for the next step after tokenizing? I couldn't find the categorizing functions so ima skip this file 👯

tests/test_main.c
```diff
@@ -3,21 +3,23 @@
...
+/*   By: sqiu <sqiu@student.42vienna.com>           +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/06/02 10:40:30 by gwolf             #+#    #+#             */
-/*   Updated: 2023/07/07 11:32:54 by gwolf            ###   ########.fr       */
+/*   Updated: 2023/07/08 19:23:00 by sqiu             ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
-#include "test.h"
+#include "../inc/test.h"
```
gwolf-011235, line 13: This should also work without the relative path - so just test.h

tests/test_main.c
```diff
@@ -3,21 +3,23 @@
...
+#include "../lib/libft/libft.h"
```
gwolf-011235, line 14: libft.h is included in test.h

valgrind-out.txt
gwolf-011235, line 1: Delete?
Maybe add a gitignore for this file

src/lexer_tok_utils.c
```diff
@@ -0,0 +1,66 @@
...
+#include "minishell.h"
+#include "lexer_tok_utils.h"
+
+/**
+ * @brief Initiate token by reserving 1 Byte of memory space.
+ * 
+ * @param data 		Struct containing the token.
+ * @return t_error	ERR_MALLOC, SUCCESS
+ */
+t_error	ft_init_tok(t_data *data)
```
gwolf-011235, line 24: I don't understand this function :D We are creating a token of size one so we can realloc it? I think realloc would create the new_size and if the old size is not null (like at the start) it would memcpy?
Could we just set token.tok to NULL?

inc/lexer_src.h
```diff
@@ -0,0 +1,38 @@
...
+
+/* ====== Definitions ====== */
+
+//# define EOF (-1)
+# define INIT_SRC_POS (-2)
+
+typedef struct s_src
+{
+	char	*buf;
+	int		buf_size;
```
gwolf-011235, line 28: I think "buf_size" is a bit confusing. I tought it had something to do with a buffer but this is the size of the input string. Maybe "input_size"? Or just "size", since src->size would be clear.

inc/lexer_src.h
```diff
@@ -0,0 +1,38 @@
...
+//# include "minishell.h"
+
+/* ====== Definitions ====== */
+
+//# define EOF (-1)
+# define INIT_SRC_POS (-2)
+
+typedef struct s_src
+{
+	char	*buf;
```
gwolf-011235, line 27: I think change "buf" to "str"

src/lexer_list.c
```diff
@@ -0,0 +1,125 @@
...
+ * @param input 	Input string to be tokenised.
+ * @return t_error 	ERR_EMPTY, ERR_MALLOC, SUCCESS
+ */
+t_error	ft_list_token(t_data *data, char *input)
+{
+	t_src		src;
+	t_error		err;
+
+	ft_init_lexer(&src, input);
+	err = ft_tokenise(&src, &data->token);
```
gwolf-011235, line 30: Although I don't really understand ft_init_tok() but if we need it: couldn't we also put it here? So we don't need to call it outside of here

src/lexer_list.c
```diff
@@ -0,0 +1,125 @@
...
+ * @param data		Overarching struct containing all meta data.
+ * @param content	Content string to be written into new node.
+ * @return t_error 	ERR_MALLOC, SUCCESS
+ */
+t_error	ft_new_node(t_data *data, char *content)
+{
+	t_tkn_list	*new;
+	t_tkn_list	*prev;
+
+	new = (t_tkn_list *)malloc(sizeof(t_tkn_list));
```
gwolf-011235, line 70: I don't think you need to cast the return of malloc to (t_tkn_list *)

src/lexer_list.c
```diff
@@ -0,0 +1,125 @@
...
+#include "lexer_utils.h"
+
+/**
+ * @brief Create a list of token.
+ * 
+ * @param data 		Overarching struct containing all meta data.
+ * @param input 	Input string to be tokenised.
+ * @return t_error 	ERR_EMPTY, ERR_MALLOC, SUCCESS
+ */
+t_error	ft_list_token(t_data *data, char *input)
```
gwolf-011235, line 24: Alternative name: ft_create_token_list()

inc/minishell.h
```diff
@@ -46,6 +51,61 @@
...
+/**
+ * @brief Token structure.
+ * 
+ * Includes token string and the token size.
+ * @param tok String containing token.
+ * @param tok_size Size of token string.
+ */
+typedef struct s_tok
+{
+	char	*tok;
```
gwolf-011235, line 65: Rename to str so tok->str

inc/minishell.h
```diff
@@ -46,6 +51,61 @@
...
+	int		tok_size;
```
gwolf-011235: Rename to size so tok->size

src/lexer_tok.c
```diff
@@ -0,0 +1,136 @@
...
+ * @param token 	Created token.
+ * @param s 		Token string.
+ * @return t_error 	ERR_MALLOC, SUCCESS
+ */
+t_error	ft_create_tok(t_tok *token, char *s)
+{
+	if (!token || !s)
+		return (ERR_EMPTY);
+	token->tok_size = ft_strlen(s);
+	token->tok = ft_realloc (token->tok, token->tok_size + 1, 1);
```
gwolf-011235: If we just malloc that, then we wouldn't need to rellaoc?

src/lexer_tok.c
```diff
@@ -0,0 +1,136 @@
...
+ * Fills buffer string until delimiters are reached.
+ * Delimiters: Space, tab, newline, pipe
+ * @param src 		Struct containing the input string,
+ * 					its length and current position.
+ * @param buf 		Buffer string to be filled.
+ * @param buf_pos 	Current position in buffer string.
+ * @param buf_size 	Buffer malloced size. 
+ * @return t_error 	ERR_EMPTY, ERR_EOF, SUCCESS
+ */
+t_error	ft_partition(t_src *src, char *buf, int *buf_pos, int *buf_size)
```
gwolf-011235: I think a problem is quotes: I there are quotes in the string then spaces shouldn't be skipped: "hello gabriel" should be one token.
So it would to check if there is a quote. If yes then add _everything_ until the closing quote is found.

src/lexer_tok.c
```diff
@@ -0,0 +1,136 @@
...
+t_error	ft_tokenise(t_src *src, t_tok *token)
+{
+	char	*buf;
+	int		buf_size;
+	int		buf_pos;
+	t_error	err;
+
+	if (!src || !src->buf || !src->buf_size)
+		return (ERR_EMPTY);
+	buf_size = 2048;
```
gwolf-011235: We could have the buffer outside of the tokenise function. Right now for every created token the buffer has to be malloc'd. If we could we reuse him, that would be fancy

tests/test_lexer.c
gwolf-011235, line 1: A good test would be 
> echo "Hello World"
Which should be two tokens

gwolf-011235 (changes requested):
- I think we can get rid of the del function pointer
- Quotes are not handled right now. Either add something before merge or we add it later?
- The fancy buffer is it's own issue i think
- Style: i don't think we should add token and token list to data since it's just temp stuff. So every tokenise function is isolated from main and gets directly passed to expand and fill cmd
-   Style: rename some lexer_ files to token_ files? I think technically tokenising is a part of lexing but I think it would make it clearer

src/lexer_tok.c
u413-284-si: True, it's still a relic of when the token itself got malloced. Omitted the free function

src/lexer_tok.c
u413-284-si: The difference lies herein: if err == ERR_EOF, then ft_create_tok is called without catching its return value (which would be SUCCESS normally). If err is sth else, the return value is caught and err overwritten. As we want to pass on the ERR_EOF to the calling function of ft_tokenise, this differentiation is necessary.

u413-284-si: Created boolean which is set to 1 when err = ERR_EOF. Boolean passed on to ft_create_tok which returns ERR_EOF after creation of token. Therefore the err value can be caught each time.

src/lexer_utils.c
u413-284-si: ft_del_content function deleted

src/lexer_utils.c
u413-284-si: Content string ist being freed directly at this point making ft_del_content obsolete.

Makefile
u413-284-si: Agreed

src/lexer_tok.c
u413-284-si: Initially this was implemented because the malloced string and its stored value were not carried on to the calling function, unless the string was previously malloced in the calling function. Somehow it works now to malloc in this subfunction and returning the pointer to its calling function; the string value is preserved.

Therefore ft_realloc is replaced with malloc.

inc/minishell.h
u413-284-si: Agreed

src/lexer_tok.c
u413-284-si: Function integrated that takes quotes into account and creates a single token of every content inside of quotes.

src/lexer_tok.c
u413-284-si: No immediate benefit, therefore in backlog

tests/test_lexer.c
u413-284-si: New test for quotes: export string="hello world" as input

gwolf-011235 (approved):
Very nice! But I will open an issue for the buffer 🗡️

PR #28, opened 2023-07-14
7 expander

### LINKED ISSUE #7: Expander
Expands some inputs

- [ ] $VAR
- [ ] $(VAR)
- [ ] $?

gwolf-011235:
### Implements expansion
1. Driver function to loop through string
- Loop through the string and look for special chars
- ~: Tilde expand
- $: Variable expand
- Quotes: removal
- Single quotes get jumped over and removed
- Double quotes turn on/off a switch and get removed

2. Tilde expansion
- Can handle "normal" tilde which gets replaced with $HOME
- ~+ gets replaced with $PWD
- ~- gtes replaced with $OLDPWD
- If tilde is quoted or not right positioned or $VAR is not found don't expand

3. Variable expansion
- Handles variable expansion
- If not found empty string is inserted
- Variable has to start with underscore or alphabetic. The underscore and alphanumeric are allowed.

### Update test_main
Test main now has some global variables which can imported into the test file via keyword "extern"

Closes #7


src/expand_tilde.c
```diff
@@ -0,0 +1,140 @@
...
+ * @param replace Where to save replace string.
+ * @param pos Current position.
+ * @return t_error SUCCESS, ERR_NOT_FOUND, ERR_MALLOC
+ */
+t_error	ft_get_tilde_replace(t_tok token, t_hashtable *symtab, t_tok *replace, size_t *pos)
+{
+	char		*target;
+	t_env_var	*env_var;
+
+	if (!ft_strcmp(token.str, "~"))
```
u413-284-si: Do we need an extra ft_strcmp()? Couldn't we just use ft_strncmp() from libft and pass it ft_strlen(str1) additionally?

src/expand_tilde.c
```diff
@@ -0,0 +1,140 @@
...
+	if (input[i] == '+')
+	{
+		token->str = "~+";
+		i++;
+	}
+	else if (input[i] == '-')
+	{
+		token->str = "~-";
+		i++;
+	}
```
u413-284-si: Could loose the brackets by putting i++ outside of the if-condition at its end, as it is performed anyways.

src/expand_tilde.c
```diff
@@ -0,0 +1,140 @@
...
+		return (SUCCESS);
+	if (err != SUCCESS)
+		return (err);
+	err = ft_insert_replace(input, *pos, token, replace);
+	free(replace.str);
+	if (err != SUCCESS)
+		return (err);
+	*pos += replace.len;
+	return (SUCCESS);
+}
```
u413-284-si: Why are  the return codes SUCCESS even when an error has occured, such as no token.str exists?

src/ft_string.c
```diff
@@ -71,3 +71,24 @@ char	*ft_strncat(char *dst, const char *src, size_t n)
...
+ */
+int	ft_strcmp(const char *str1, const char *str2)
+{
+	while (*str1 && *str1 == *str2)
+	{
+		str1++;
+		str2++;
+	}
+	return ((int)(unsigned char)(*str1) - (int)(unsigned char)(*str2));
+}
```
u413-284-si: As mentioned before, not sure we need this function as libft already provides ft_strncmp()

src/expand_var.c
```diff
@@ -0,0 +1,130 @@
...
+	t_tok	replace;
+	t_error	err;
+
+	if ((*input)[*pos + 1] == '0' || (*input)[*pos + 1] == '?')
+		err = ft_special_var((*input)[*pos + 1], &replace, &token, info);
+	else
+	{
+		err = ft_get_var_token(*input, *pos, &token);
+		if (token.len == 0)
+			return (SUCCESS);
```
u413-284-si: So it returns SUCCESS even if the variable has not been found?

src/expander.c
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * @brief Search for token in input string and replace with replace.
+ *
+ * @param input String.
+ * @param pos Current position.
+ * @param token Struct containing the searched for token.
+ * @param replace Struct containing the replacement.
+ * @return t_error SUCCESS, ERR_MALLOC
+ */
+t_error	ft_insert_replace(char **input, size_t pos, t_tok token, t_tok replace)
```
u413-284-si: Rly like this function, kudos to that!

src/expander.c
```diff
@@ -0,0 +1,140 @@
...
+{
+	char	*str1;
+	char	*str2;
+
+	str1 = input + pos;
+	str2 = str1 + 1;
+	while (*str1)
+	{
+		*str1++ = *str2++;
+	}
```
u413-284-si: Could omit the brackets

src/expander.c
```diff
@@ -0,0 +1,140 @@
...
+	return (SUCCESS);
+}
+
+/**
+ * @brief Handle single quotes.
```
u413-284-si: Could be more detailed on what happens with single quotes. In such describing that it keeps everything inside as is and moves on to the end of the single quotes to look further for stuff to be expanded.

src/expander.c
```diff
@@ -0,0 +1,140 @@
...
+{
+	ft_eat_char(expr, *pos);
+	while (expr[*pos] != '\'')
+		(*pos)++;
+	ft_eat_char(expr, *pos);
+	return (SUCCESS);
+}
+
+/**
+ * @brief Handle double quotes.
```
u413-284-si: Could be more detailed on what happens with double quotes.

tests/test_env_envp.c
```diff
@@ -6,14 +6,14 @@
...
 /* ************************************************************************** */
 
 #include "test.h"
 
-t_hashtable	*symtab;
-char		**envp;
+extern t_hashtable	*symtab;
+extern char			**envp;
 
 void	test_env_envp_two(void)
```
u413-284-si, line 18: What is the functionality of these two functions?

tests/test_env_envp.c
```diff
@@ -31,15 +31,15 @@ void	test_env_envp_two(void)
...
 
 void	test_env_envp_empty(void)
 {
 	ft_envp_create(symtab, &envp);
 	printf("This is envp[0]: %p\n", envp[0]);
-	ft_envp_destroy(envp);
+	ft_envp_destroy(&envp);
 }
 
 void	test_env_envp(void)
```
u413-284-si, line 45: Que es eso?

u413-284-si (changes requested):
If I understand correctly, the expander requires the token string as input parameter. The lexer provides a list of token. 

Therefore we need an overarching interface function between the lexer and the expander which extracts the token string out of the individual nodes, passes it on to the expander, then gives the result on to the parser. And iterates through the list of token.

src/expand_tilde.c
gwolf-011235: I think it's neater to read, since it doesn't involve another function call inside the function. 
Also it doesn't need ft_strlen() since it's running as long as the string isn't terminated. That means for every call of strcmp() we save a strlen() which means a speed increase of 100%!

src/expand_tilde.c
gwolf-011235: Not quite: i only want to i++ if i find a plus '+' or a minus '-' after the tilde. If there is neither of the two (which is perfectly valid) i don't want i++
I also tried to somehow get rid of the double i++. But since I'm also assigning the correct token I need to know which char exactly i found. I think the alternative would be to check either plus or minus and the check again which one of the two.

src/expand_tilde.c
gwolf-011235: If a tilde is on the wrong position it doesn't get expanded. That's not an error per se: "hel~lo" is a valid string, it just means the tilde won't get expanded. 
Same case with no token found:  if you write ~. or ~* we don't (shouldn't) expand.
So it's still SUCCESS, even though no expansion happened.

src/ft_string.c
gwolf-011235: See above - I think it's neat :D

src/expand_var.c
gwolf-011235: Yeah: if the token.len is zero it means after the $ sign there were no valid chars. THis means we don't expand the $ and also have to print it. Therefore I want to exit the function. Kinda the same reasoning like with tilde replace

src/expander.c
gwolf-011235: :heart:

src/expander.c
gwolf-011235: True. 
I like to have them when I don't need the space/lines, but I also don't need them if it's confusing

src/expander.c
gwolf-011235: True. I just wanted to push :D I'll amend

src/expander.c
gwolf-011235: True again. I'll ammend

gwolf-011235 (commented):
I would leave strcmp().
In exchange I'll remove the bracktes :D
I'm gonna amend the doxy comments.

src/expand_tilde.c
u413-284-si: I am quite positive that ft_strlen() does not require that many resources that we would notice any speed difference. I am an advocate of not reinventing the wheel and utilise all the functions which are already at our disposition, but here's a deal:

You get to use your ft_strcmp() whilst I get to keep my buffer creation inside my tokeniser.

src/expand_tilde.c
u413-284-si:
```
if (input[i] == '+')
    token->str = "~+";
else if (input[i] == '-')
    token->str = "~-";
if (input[i] == '+' || input[i] == '-')
    i++;
```

Like this you could save 4 lines.

src/expand_tilde.c
u413-284-si: To my understanding a SUCCESS return only is justified, if the function has performed its role, in this case "expanded the tilde". If it has not done this, then I'd consider this an error in the sense that it could not do its job. Even though it might not be a fatal error causing the disruption of the whole program, I would call this sth like ERR_NOEXPAND.

src/ft_string.c
u413-284-si: See proposal ;)

src/expand_var.c
u413-284-si: Do we print the $ in this case? I don't see it anywhere happening in the code...

src/expand_tilde.c
gwolf-011235: I gonna delete the strcmp 🤣

src/expand_tilde.c
gwolf-011235: True! I think I was trying to avoid double checking the char. Changed

src/expand_tilde.c
gwolf-011235: You're right, it does make it more clearer - added the error code and handling in functions.

src/expand_var.c
gwolf-011235: Good catch! I thought I added a similiar handler like in tilde, but forgot. Added a pos jumper.
This comment also lead me to another bug: I checked the first char, but then while loop regardless. Fixed.

I also found two additional problems (not quite sure how to handle them yet): 
- If the preeceding token is << (infile) the var gets not expanded so << $test has $test as delimiter. I think a fix for this depends on our classification system of the tokens
- If the string is quoted, but the $ is not, the dollar sign gets removed (even though the var doesn't start with a valid char). I don't quite understand why. This means $"test" results in test, whereas "$"test results in $test

src/expand_tilde.c
u413-284-si: 😂

src/expander.c
gwolf-011235: Removed.

src/expander.c
gwolf-011235: Added better comment

src/expander.c
gwolf-011235: Added better comment

tests/test_env_envp.c
gwolf-011235: With the extern keyword i can "import" the globals from main. The main has some vars globally defined, which can be used in the testfile. I like to use it for t_hashtable *symtab for example. Then I don't need to pass the pointer around, I can just use it.

tests/test_env_envp.c
gwolf-011235: I recompiled the test file with tclean and found out that i didn't update an old test: ft_envp_destroy() is from another module. I just removed the ampersand to not pass a ***

src/expand_var.c
u413-284-si: 1. Yes, in that case we would save the string as a delimiter in our cmd struct. But as your expander runs before the parser, you would need to incorporate a check for if input[i] = '$' && input[i-1] = '<' && input[i-2] = '<' and jump over the expander in this case.
2. Another special case to be implemented...

tests/test_env_envp.c
u413-284-si: I was referring to the functions:

void	test_env_envp_two(void)
void	test_env_envp_empty(void)

I guess these are to create the environment, one with two variables and one which is empty

src/expand_tilde.c
gwolf-011235: ft_strcmp has been completely eradicated 💇

src/ft_string.c
gwolf-011235: Strcmp is no mo

src/expand_var.c
gwolf-011235: Second one is fixed: i refactored to also pass if the $ is in quotes. If it is not in quotes and the next char is a quote then the $ gets deleted. I still don't exactly know why this should be the correct behavior.
Added testcases for double and single quote variations

tests/test_env_envp.c
gwolf-011235: Yeah these are testcases for creation of envp: two vars in environment and empty environment.

src/expand_tilde.c
u413-284-si: wuhuuuuuuuuuuuuuuu

src/expand_var.c
u413-284-si: we also good with the first issue?

src/lexer_expand.c
gwolf-011235, line 1: This function expands a tkn_list. It acts as a connector between the expand-module and the list-module. In the .h file is the prototype of the expand_expr function. This way everything in the expand.h can be kept separted from the rest of the program.

If it sees a heredoc token it jumps over the token, and the next token (which would be the heredoc delimiter which doesn't expand).
Else it calls ft_expand_expr() on token->content

gwolf-011235 (commented):
The new "bridge" function connetcs the two modules which makes it possible to isolate the expand.h.
This functions expands all tokens except heredocs.

The test_main also got an update with a setup and teardown function for the global g_data.

Everything should be fix now.

PR #30, opened 2023-07-18
29 have redirects as individual tokens

### LINKED ISSUE #29: Have redirects as individual tokens
< infile currently will be saved as two tokens: '<' and 'infile'
<infile shall have the same result.

The same goes for >outfile, <<delimiter, >>outfile (in append mode)

Concerns the lexer.

u413-284-si:
redirects will be saved individually as token non-regarding if there is a space after the redirect or not:

<infile and < infile will become two tokens: '<' and 'infile'

<< will become two tokens: '<' and '<'

closes issue #29


src/lexer_tok.c
```diff
@@ -107,15 +107,16 @@ t_error	ft_partition(t_src *src, t_buf *tmp)
...
-		else if ((c == '\n' || c == '|') && tmp->cur_pos > 0)
+		else if ((c == '\n' || c == '|' || c == '<' || c == '>')
+			&& tmp->cur_pos > 0)
 		{
 			ft_unget_char(src);
 			break ;
 		}
 		else
 		{
 			ft_add_to_buf(c, tmp);
```
gwolf-011235: If you pass here t_src you could - in case of redirect - peek if the next char is the same as c. If yes call get_char and also add it to buf.

src/lexer_tok.c
```diff
@@ -107,15 +107,16 @@ t_error	ft_partition(t_src *src, t_buf *tmp)
 			ft_add_quoted_str(c, src, tmp);
 		else if ((c == ' ' || c == '\t') && tmp->cur_pos > 0)
 			break ;
-		else if ((c == '\n' || c == '|') && tmp->cur_pos > 0)
+		else if ((c == '\n' || c == '|' || c == '<' || c == '>')
+			&& tmp->cur_pos > 0)
 		{
 			ft_unget_char(src);
```
gwolf-011235: Unget_char could return SUCCESS. Then you could return it, saving you 3 lines :D

tests/test_lexer.c
```diff
@@ -25,7 +25,7 @@ void	test_lexer(void)
 		ft_exit_failure(data, ERR_MALLOC);
 	data->lst_head = NULL;
 
-	printf("\nSimple command: ls\n");
+/* 	printf("\nSimple command: ls\n");
```
gwolf-011235: I think you could comment them back in, so the next time every test gets run again

src/lexer_tok.c
```diff
@@ -107,15 +107,16 @@ t_error	ft_partition(t_src *src, t_buf *tmp)
 			ft_add_quoted_str(c, src, tmp);
 		else if ((c == ' ' || c == '\t') && tmp->cur_pos > 0)
 			break ;
-		else if ((c == '\n' || c == '|') && tmp->cur_pos > 0)
+		else if ((c == '\n' || c == '|' || c == '<' || c == '>')
```
gwolf-011235: You could also use ft_strchr() to search the string "\n|<>" for the occurrence of c to shorten this a bit

gwolf-011235 (commented):
I think it makes sense to modify ft_add_to_buf. I think it would make the code more readable later on, if we just have one token for >> and << .

src/lexer_tok.c
u413-284-si: what i could do is write another function who - as you proposed - receives t_src instead of ft_add_to_buf(). In case the next char is the same redirect (use ft_peek_char), then add it as well to buf, if not just add the current char.

src/lexer_tok.c
u413-284-si: you ingenious bastard you!

tests/test_lexer.c
u413-284-si: yeah, will do

src/lexer_tok.c
u413-284-si: not bad, not bad at all. haben wir gestern unser Zielwasser getrunken ;D

src/lexer_tok.c
u413-284-si: Added ft_add_double_redirect(t_src *src, char *c, t_buf *tmp) to create one token out of << or >> respectively.

src/lexer_tok.c
```diff
@@ -107,16 +101,13 @@ t_error	ft_partition(t_src *src, t_buf *tmp)
...
-			break ;
-		}
+		else if (ft_strchr("\n|<>", c) && tmp->cur_pos > 0)
+			return (ft_unget_char(src));
 		else
 		{
 			ft_add_to_buf(c, tmp);
-			if (c == '\n' || c == '|')
-				break ;
+			if (ft_strchr("\n|<>", c))
```
gwolf-011235, line 107: Maybe I'm missing something but I don't think we need to check for newline and pipe?
If it's just two characters I prefer the explicit way of checking (i.e. c == '<') since it's easier to read. It also needs no function call.

src/lexer_tok_utils.c
```diff
@@ -56,3 +56,19 @@ t_error	ft_add_to_buf(char c, t_buf *tmp)
...
+
+t_error	ft_add_double_redirect(t_src *src, char *c, t_buf *tmp)
+{
+	char	next;
+	t_error	err;
+
+	err = ft_peek_char(src, &next);
+	if (err != SUCCESS)
+		return (err);
+	if (ft_strchr("<>", *c) && next == *c)
```
gwolf-011235, line 76: Since we only get into this function after checking for < and > we don't need to check the char again. So next == *c should be sufficient.

src/lexer_tok.c
gwolf-011235, line 1: I just saw now that after the addition of t_buf some of the doxygen comments are outdated. They still refer to params like buf_size which are now contained in the struct.

Also: i would rename the buffer from tmp to buf or something. I know the buffer is temporaray but its still used for a while. For example: In the realloc part of ft_add_to_buf() it's confusing having t_buf tmp and char * temp

src/lexer_tok.c
```diff
@@ -75,10 +72,7 @@ t_error	ft_tokenise(t_src *src, t_tok *token)
 	if (tmp.cur_pos >= tmp.size)
 		tmp.cur_pos--;
 	tmp.str[tmp.cur_pos] = '\0';
-	eof = 0;
-	if (err == ERR_EOF)
-		eof = 1;
-	err = ft_create_tok(token, tmp.str, eof);
+	err = ft_create_tok(token, tmp.str);
```
gwolf-011235: Why do we not need the bool anymore? Is it becaue of ERR_OUT_OF_BOUNDS?

src/lexer_tok_utils.c
```diff
@@ -56,3 +56,19 @@ t_error	ft_add_to_buf(char c, t_buf *tmp)
...
 	return (SUCCESS);
 }
+
+t_error	ft_add_double_redirect(t_src *src, char *c, t_buf *tmp)
+{
+	char	next;
+	t_error	err;
+
+	err = ft_peek_char(src, &next);
+	if (err != SUCCESS)
```
gwolf-011235, line 74: Just for clarification: This would only fire if we are - why ever - on position 0 in the src string right? So it's like checking for empty vars although we always pass vars

gwolf-011235 (commented):
I think if the testcase is ok, then it should work w/o newline.
I think add_double_redirect can be improved

src/lexer_tok.c
gwolf-011235: We still have to check for \n and | to break out of loop

src/lexer_tok_utils.c
gwolf-011235: No since we also check for \n and |

src/lexer_tok.c
gwolf-011235: Nice rename

PR #33, opened 2023-07-21
feat: add ft_handle_input

### LINKED ISSUE #32: Create input handler
Driver function that encompasses:

syntax check
lexer
expander
parser

returns cmd list to executor

u413-284-si:
Driver function that is called by main and performs in linear manner following functionalities:

1. syntax check
2. lexer
3. expander
4. parser
5. executor

closes #32


gwolf-011235 (approved):
Looks very good, nice work

PR #38, opened 2023-07-25
6 parser

### LINKED ISSUE #6: Parser
- [ ] Split input nach pipes (ft_split?)
- [ ] Command table init
- [ ] Token stuff

u413-284-si:
Implements a parser to go through the token list and create a list of commands by assigning each token into following categories:

- exe
- args
- infile
- outfile
- heredoc delimiters
- append mode

1. exe: If the bool exe_found has not been set yet, the first string encountered which does not belong to a redirect is set as the executable.
2. args: When the bool exe_found is set, any string which does not belong to a redirect is set as arg.
3. When a redirect indicating an infile (<) is encountered, the following string token is read and opened as infile and its file descriptor saved. In case of several infiles, the last infile will be retained and the previous abandoned.
4. When a redirect indicating an outfile (>) is encountered, the following string token is read and opened as outfile and its file descriptor saved. In case of several outfiles, the last outfile will be retained and the previous abandoned.
5. When a redirect indicating a heredoc  (<<) is encountered, the following string token is read and saved as delimiter to that heredoc. In case of several heredocs, all delimiters are retained in their order of appearance.
6. When a redirect indicating an outfile opened in append mode (>>) is encountered, the following string token is read and opened in append mode and its file descriptor saved as well as a bool indicating append mode is set. In case of several outfiles, the last outfile will be retained and the previous abandoned.

The input handler is adapted to pass on the required parameters to the parser and receives the created cmd list which can be passed unto the executor.

A set of tests verifies the correct behaviour of the parser:

- 	test_one_cmd();
- 	test_two_cmds();
- 	test_five_cmds();
- 	test_three_infiles();
- 	test_heredoc();
- 	test_three_heredocs();
- 	test_three_outfiles();
- 	test_append();

Closes #6


inc/test.h
```diff
@@ -17,7 +17,6 @@
 # include "mod_env.h"
 # include "mod_expand.h"
 # include "mod_handle_input.h"
-# include "mod_lexer.h"
```
gwolf-011235, line 20: Why deleted ?

src/lexer.c
```diff
@@ -1,109 +1,130 @@
...
- * 
- * 		<infile exe opt1 opt2 opt3 >outfile
- * 
- * Each of these input strings represent a token. These tokens are saved
- * according to each simple command. 
- * 
- * @param data 	Overarching struct to hold all necessarry data.
- * @param input Input string delivered by <readline>.
- * @return t_err 
+ * @brief Create a list of token.
```
gwolf-011235, line 24: Maybe add a little more info what it does

src/lexer.c
```diff
@@ -1,109 +1,130 @@
...
-		if (input[i + 1] == '>')
-			data->err = ft_save_append(data, input);
-		else
-			data->err = ft_save_outfile(data, input);
-	}
-	else
-		data->err = ft_categorise2(data, input, exe_found);
-	if (data->err != SUCCESS)
-		return (data->err);
+	new = (t_tkn_list *)malloc(sizeof(t_tkn_list));
```
gwolf-011235: I think it's recommended not to cast the return of malloc()

src/lexer.c
```diff
@@ -1,109 +1,130 @@
...
 	return (SUCCESS);
 }
 
-
-t_err	ft_categorise2(t_data *data, char *input, bool *exe_found)
+/**
+ * @brief Iterate through token list and free all nodes.
+ *
+ * @param lst Token list to be freed.
+ * @param del Pointer to function used for clean up.
```
gwolf-011235: Can be rm'd

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+/*   By: sqiu <sqiu@student.42vienna.com>           +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2023/06/23 13:13:28 by sqiu              #+#    #+#             */
+/*   Updated: 2023/07/26 01:25:52 by sqiu             ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+/**
+ * @file parser.c
+ * @brief
```
gwolf-011235: What it do :0

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+ */
+
+#include "mod_parser.h"
+
+/**
+ * @brief 			Driver function to parse token list.
+ * 
+ * Creates cmd structures and fills them with the token list.
+ * @param lst 		Pointer to list of tokens.
+ * @param cmd 		Pointer pointer to first node of cmd list.
```
gwolf-011235: Rename to cmd_head to clarify?

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+	return (new);
+}
+
+/**
+ * @brief Initiates the cmd struct.
+ * 
+ * @param new Newly created cmd struct.
+ * @return t_err 	ERR_MALLOC, SUCCESS
+ */
+t_err	ft_init_cmd(t_cmd **new)
```
gwolf-011235: I suggest ft_create_cmd() or ft_create_init_cmd(). 
For me a sole "init" implies that it just initializes something already created

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+ */
+t_err	ft_categorise(t_tkn_list **lst, t_cmd *new, bool *exe_found,
+	bool *cmd_complete)
+{
+	t_tkn_list	*tmp;
+	t_err		err;
+
+	err = SUCCESS;
+	tmp = *lst;
+	if (ft_strncmp(tmp->content, "<", ft_strlen(tmp->content)) == 0)
```
gwolf-011235: If tmp->content is an empty string (for example after expansion of an empty environmental variable) ft_strlen returns zero, which in turn would make this comparison always true. Take ft_strlen of "<" instead.
Concerns all ft_strncmp in this function

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+	int			fd_in;
+	t_tkn_list	*tmp;
+
+	tmp = *lst;
+	tmp = tmp->next;
+	new->infile = 1;
+	if (new->fd_in > 0)
+		close(new->fd_in);
+	fd_in = open(tmp->content, O_RDONLY);
+/* 	if (fd_in < 3)
```
gwolf-011235: Why commented out? Shouldn't a error check for == -1 suffice? I think if you close for example fd 1, then it can get reassigned, so only -1 is error.

Or is it commented out since all infiles are opened regardless of error? Or am i misremembering something :D

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+	if (!new->delim_buf)
+	{
+		new->delim_buf = malloc(ft_strlen(tmp->content) + 1);
+		if (!new->delim_buf)
+			return (ERR_MALLOC);
+		ft_strlcpy(new->delim_buf, tmp->content, ft_strlen(tmp->content) + 1);
+	}
+	else
+	{
+		buf = ft_strjoin(new->delim_buf, " ");
```
gwolf-011235: I think a problem with this approach is spaces in the delimiter. It's possible to choose as heredoc delim "hello world". In this case it will read until hello world is entered (the quotes are removed beforehand. 
Alternative: put the delims in a linked list.

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+{
+	t_tkn_list	*tmp;
+	char		*buf;
+	char		*buf2;
+
+	tmp = *lst;
+	tmp = tmp->next;
+	if (!new->delim_buf)
+	{
+		new->delim_buf = malloc(ft_strlen(tmp->content) + 1);
```
gwolf-011235: I think a call to ft_strdup() would shorten this a bit.

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+t_err	ft_save_infile(t_tkn_list **lst, t_cmd *new)
+{
+	int			fd_in;
+	t_tkn_list	*tmp;
+
+	tmp = *lst;
+	tmp = tmp->next;
+	new->infile = 1;
+	if (new->fd_in > 0)
+		close(new->fd_in);
```
gwolf-011235: Technically we should also check the return of close. But I'm not quite sure what happens if it fails :D

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+	tmp = *lst;
+	tmp = tmp->next;
+	new->outfile = 1;
+	if (new->fd_out)
+		close(new->fd_out);
+	if (append)
+		fd_out = open(tmp->content, O_WRONLY | O_APPEND | O_CREAT, 0644);
+	else
+		fd_out = open(tmp->content, O_RDWR | O_TRUNC | O_CREAT, 0644);
+	if (fd_out < 3)
```
gwolf-011235: Same as above: shouldn't it be "== -1" instead of < 3

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+ * @param lst 			List of token.
+ * @param new 			New cmd to be filled.
+ * @param exe_found 	Bool to indicate if executable has been found.
+ * @return t_err 		ERR_MALLOC, SUCCESS
+ */
+t_err	ft_save_exe(t_tkn_list *lst, t_cmd *new, bool *exe_found)
+{
+	int	len;
+
+	len = ft_strlen(lst->content);
```
gwolf-011235: You can save some lines with ft_strdup()

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+	if (!new->arg_buf)
+	{
+		new->arg_buf = malloc(len + 1);
+		if (!new->arg_buf)
+			return (ERR_MALLOC);
+		ft_strlcpy(new->arg_buf, lst->content, len + 1);
+	}
+	else
+	{
+		tmp = ft_strjoin(new->arg_buf, " ");
```
gwolf-011235: Same problem as above with space as a delimiter between args: if the arg is "Hello World" it would be read as two args (the quotes are removed beforehand). 
Linked list or alloc an array of five. If it's to small realloc?

src/parser_utils.c
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @brief Finish up last cmd in cmd_list.
+ * 
+ * @param new 		Last cmd.
+ * @param cmd 		List of cmds.
+ * @return t_err 	ERR_MALLOC, SUCCESS
+ */
+t_err	ft_finish_cmd_list(t_cmd *new, t_cmd **cmd)
+{
+	new->args = ft_split(new->arg_buf, ' ');
```
gwolf-011235: ft_split() sadly not usable - see above :(

tests/infile
```diff
@@ -0,0 +1 @@
+yomomma is beautiful
```
gwolf-011235, line 1: Yo mama's so wise, Yoda often texts her for advice.

tests/test_expand_expr.c
```diff
@@ -18,7 +18,7 @@ extern t_hashtable	*g_symtab;
 extern t_info		g_info;
 extern int			g_err_count;
 
-int	test_wrapper(char *testname, char *test, char *expect)
+static int	test_wrapper(char *testname, char *test, char *expect)
```
gwolf-011235, line 21: I had the same problem of redefinition :D I recently used test_ft_name() so it's clear which function is tested

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+ * @return t_cmd* 
+ */
+t_cmd	*ft_new_cmd(t_cmd *curr, t_cmd **cmd, bool *exe_found,
+	bool *cmd_complete)
+{
+	t_cmd	*new;
+	t_err	err;
+
+	new = NULL;
+	curr->args = ft_split(curr->arg_buf, ' ');
```
gwolf-011235: ft_split() doesn't work, see down below. Problem are args which have spaces inbetween

src/parser.c
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * @brief Adds current cmd to cmd_list and creates a new cmd struct.
+ * 
+ * @param curr 			Current cmd struct.
+ * @param cmd 			List of cmd structs.
+ * @param exe_found 	Bool to indicate if executable has been found.
+ * @param cmd_complete 	Bool to indicate if cmd ist complete.
+ * @return t_cmd* 
+ */
+t_cmd	*ft_new_cmd(t_cmd *curr, t_cmd **cmd, bool *exe_found,
```
gwolf-011235: I would call it ft_next_cmd() since this only gets called after the first cmd has been filled

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,168 @@
...
+		close(new->fd_out);
+	if (append)
+		fd_out = open(tmp->content, O_WRONLY | O_APPEND | O_CREAT, 0644);
+	else
+		fd_out = open(tmp->content, O_RDWR | O_TRUNC | O_CREAT, 0644);
+	if (fd_out < 3)
+		return (ERR_BAD_FD);
+	new->fd_out = fd_out;
+	if (append)
+		new->append = 1;
```
gwolf-011235, line 97: Do we even need this bool in the cmd struct? Since append mode is already activated through open, I don't think we care anymore

gwolf-011235 (changes requested):
I think the solution via ft_strjoin() and ft_split() to collect the args / heredocs breaks as soon as somebody has an arg or a delim with spaces. The quotes are removed by then, so we can't rely on them (I also don't think it's wise to keep them and remove later. I think this would introduce other problems).

A solution could be linked list of char pointers. Or - and I like this more - we always allocate an array of like 5 char *. If it's not enough double it. If we have less: who cares :D We could set is as a macro in the config.h (STD_ARGV_SIZE)

Maybe I'm missing something but I would also get rid of the splitting between exe and args. Execve() receives an argv array, where by convention arg[0] contains the filename (= exe). I followed this convention in the builtins. As args I always check after the zero'th, since this would be the program name itself. 
So for executing we have to combine them anyway. Therefore if we already build an array while filling the cmd, we wouldn't need to combine them afterwards.

inc/test.h
u413-284-si: Cause it's unnecessary, the test_lexer.c only requires mod_handle_input.h

src/lexer.c
u413-284-si: That's basically it :D what else should there be

src/lexer.c
u413-284-si: why not? that's actually how the official definition is afaik. We just leave it away out of convenience as it is implicitly assumed by the compiler

src/lexer.c
u413-284-si: rm?

src/parser.c
u413-284-si: upsi, forgot that one apparently

tests/test_expand_expr.c
u413-284-si: hmmm? but isn't that a general function which is used by all tests?? show me what you mean

src/parser.c
u413-284-si: How about assigning a different IFS? a greek semicolon for all I care

src/parser.c
u413-284-si: But it's newly created O.O

src/parser_fill_cmd.c
u413-284-si: This might come in handy later on, for now I primarily used it for the test which checks with that bool whether the according output shall be tested and printed

u413-284-si:
> I think the solution via ft_strjoin() and ft_split() to collect the args / heredocs breaks as soon as somebody has an arg or a delim with spaces. The quotes are removed by then, so we can't rely on them (I also don't think it's wise to keep them and remove later. I think this would introduce other problems).
> 
> A solution could be linked list of char pointers. Or - and I like this more - we always allocate an array of like 5 char *. If it's not enough double it. If we have less: who cares :D We could set is as a macro in the config.h (STD_ARGV_SIZE)
> 
> Maybe I'm missing something but I would also get rid of the splitting between exe and args. Execve() receives an argv array, where by convention arg[0] contains the filename (= exe). I followed this convention in the builtins. As args I always check after the zero'th, since this would be the program name itself. So for executing we have to combine them anyway. Therefore if we already build an array while filling the cmd, we wouldn't need to combine them afterwards.

As mentioned above another simple solution would be to replace the spaces with a non-printable char, a vertical tab for example which should never appear in an arg.

The exe may be the first element of the arg**, but during the executing process it is necessary to assign the right path to the respective exes. Then this complete executable (= path + exe) can be assigned to the arg[0]. This process already has been implemented in pipex and can be used without hesitation.

src/parser.c
u413-284-si: done

src/parser.c
u413-284-si: It is debatable as new has already been declared in the calling function, but if it gives you inner peace :D let's go for ft_create_cmd() then

src/parser.c
u413-284-si: Wait a minute, don't we expand now before lexing? Why should there be an empty string inside of a token?

src/parser_fill_cmd.c
u413-284-si: It is commented out due to the bug of infiles not being consistently opened as they should be. If active this would abort the whole programm and hinder the testing. Needs to be addressed after having found the root cause of the infiles not being opened reliably. Furthermore, why should anyone close any fd 0, 1 or 2? We don't in our programm and it just shouldnt happen if you wanna communicate with shell

src/parser_fill_cmd.c
u413-284-si: addressed in comment below

inc/test.h
gwolf-011235: Checks out!

src/parser_fill_cmd.c
u413-284-si: oha, very true!

src/lexer.c
gwolf-011235: I think the while loop (err != EOF || !*lst_head) could be explained. Why are we looping if we have no head? Wouldn't this also loop if ft_tokenize() failed?

src/parser_fill_cmd.c
u413-284-si: It should return an error as well and lead to program abortion. Implemented it

src/lexer.c
gwolf-011235: https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc 

I like this one for answers. They basically boil down to it's not necessary and cluttering. 
But there are also some peeps which prefer to cast.
For my money it should be consistent. Since we generally don't cast the return, I think it's confusing to cast it here.

src/parser_fill_cmd.c
u413-284-si: noooooooooooooooooooooooooooooooooooooooooooooes

src/lexer.c
gwolf-011235: Removed :D

src/parser_fill_cmd.c
u413-284-si: black magic!

src/parser_fill_cmd.c
u413-284-si: see below

src/parser_utils.c
u413-284-si: see below

tests/test_expand_expr.c
gwolf-011235: I think you made this one static (in test_expand), since you also called yours test_wrapper? 
In builtin test I called mine test_wrapper at first, got a redefinition error and then just called it exec_ft_echo() - not "test_.." as I wrote. 
So a function which executes a function to be tested has the form exec_ft_name(). For example if you want to test subfunctions of your module. But in the end - I don't think it matters :D

src/parser_fill_cmd.c
gwolf-011235: Seems reasonable

src/lexer.c
u413-284-si: It loops even w/o head so that a node gets created even if the first token is directly at the end of file (no newline). err would be ERR_EOF. Without the head condition that token would not be saved in the list. If ft_tokenise() fails, it would loop but break out with an error with ft_new_node()

src/lexer.c
u413-284-si: Further down in the same page there is this which I find quite appealing :)

As with assertions, annotations that reveal intent decrease bugs. It forces an extra cognitive cross-check. 

But I agree that we should keep it consistent. As it would be more laborous to alter all instances where the cast has been omitted, we will just leave it away here.

src/parser.c
gwolf-011235: I have to admit I don't think that's very elegant. Though it probably won't matter in the scope of the subject and it's an easy fix, so yeah go for it.
Maybe use a macro so we can change it easily if we encounter a problem with a specific separator?

src/lexer.c
u413-284-si: porqueeeeeeeeeeee

src/parser.c
gwolf-011235: True! But only after the first one has already been created. We even pass the old one. 

For me a ft_new_cmd only creates a new one. 
But this function finishes the first and then creates the ft_next_cmd ;)

tests/test_expand_expr.c
u413-284-si: Gotta see your code to see what you mean. As the static solutions works just fine, let's leave it like that

src/parser.c
u413-284-si: You will see, I will choose a beautiful one ;)

src/lexer.c
gwolf-011235: I'd suggest add a shorter version of this to the doxy? Like:
Create token with ft_tokenize(). 
At least one node gets created, since lst_head is empty on first run. 
Create tokens in loop until EOF

src/parser.c
gwolf-011235: You mean parsing? :D 
Yeah and thats why an empty string can occur: "hello $mum" gets lexed into two tokens 
-  hello
-  $mum

If $mum is not set it still gets expanded resulting in an empty string (which is bash behavior). So the two token would be 
- hello 
- "" (empty string)

Now you miss my ft_strcmp :0

gwolf-011235: Oh! Or did you mean lexing (= which is tokenize?) 
Right now we lex->expand. I still need to implement field splitting which would not affect the problem: since an empty string can't be split it would remain as such.
We could implement a cleaner function which would search for empty strings and delete them from the tkn_lst.

src/parser_fill_cmd.c
gwolf-011235: Ah ok!

Why close fd 2? That's like asking why would anyone open two heredocs :D There are some crazy mofos out there :0 
I would just stick to the convention check for -1 as error. I have not (yet?) seen the check for fd < 3

src/parser.c
u413-284-si: Yeah I mean lexing, so have everything expanded before tokenising, so that the actual input gets saved in the token list. Is there any argument against that order?

src/parser.c
u413-284-si: We could call it ft_lock_and_load_cmd() :DDDDDDDDDDDDD

src/lexer.c
gwolf-011235: The doxygen has the old del function pointer as  parameter. It can be removed :)

src/parser.c
gwolf-011235: I wanted to implement it as next step so the sequence would be
Lexing - Expand (excl quote removal) - Field splitting (if Expand happened) - Quote Removal 

If Lexing is before expand it's easier for here-doc delims (they can be skipped). 
But this could also be an extra check in the expand step.

But more importantly: echo needs the empty strings. If you echo "hello $VAR world" and VAR is not set echo prints "hello  world" (two spaces) since it prints a space after each arg if there is another one.

gwolf-011235:
> > I think the solution via ft_strjoin() and ft_split() to collect the args / heredocs breaks as soon as somebody has an arg or a delim with spaces. The quotes are removed by then, so we can't rely on them (I also don't think it's wise to keep them and remove later. I think this would introduce other problems).
> > 
> > A solution could be linked list of char pointers. Or - and I like this more - we always allocate an array of like 5 char *. If it's not enough double it. If we have less: who cares :D We could set is as a macro in the config.h (STD_ARGV_SIZE)
> > 
> > Maybe I'm missing something but I would also get rid of the splitting between exe and args. Execve() receives an argv array, where by convention arg[0] contains the filename (= exe). I followed this convention in the builtins. As args I always check after the zero'th, since this would be the program name itself. So for executing we have to combine them anyway. Therefore if we already build an array while filling the cmd, we wouldn't need to combine them afterwards.
> 
> As mentioned above another simple solution would be to replace the spaces with a non-printable char, a vertical tab for example which should never appear in an arg.
> 
> The exe may be the first element of the arg**, but during the executing process it is necessary to assign the right path to the respective exes. Then this complete executable (= path + exe) can be assigned to the arg[0]. This process already has been implemented in pipex and can be used without hesitation.


Guys on the campus would also advise against using ft_split().
Problem: if somebody copies text from the web it could include a vertical tab (or any other char) :0

Idea from them: If we count the args beforehand and then malloc enough space to save them.

And re exe / args: exactly because the first arg in the end needs complete path it's easier to already have them together no? 
Arg[0] = "ls"
Then append path, free the old and after that 
Arg[0] = "/path/ls"
Else we would need a new array to store exe and args together after appending path

tests/test_expand_expr.c
gwolf-011235: Yeah they work! I just found it funny that I didn't think of that (just make it static :D)

src/lexer.c
gwolf-011235: Nice comment!

u413-284-si: Comment elaborated

src/parser_fill_cmd.c
u413-284-si: those people should not be let near any PCs. But to be foolproof, changed to  -1

src/parser.c
u413-284-si: implemented

u413-284-si:
Lexer: 
While creating the token list, each node is attributed a type indicating its functionality later on for the parser and eventually the executor. Types:
	ARG,
	INFILE,
	HEREDOC,
	OUTFILE,
	APPEND,
	PIPE,
	NEWLINE

Parser:
- No more differentiation between exe and args -> both are saved in the same char **args whereas the exe is saved in args[0].
- Size of char **args is malloced beforehand by counting the number of args including exe; same goes for delimiters and their char **delims
- Counting and categorising is performed by checking the token type assigned before in the lexer

src/parser_fill_cmd.c
```diff
@@ -0,0 +1,123 @@
...
+ */
+t_err	ft_save_infile(t_tkn_list **lst, t_cmd *new)
+{
+	int			fd_in;
+	t_tkn_list	*tmp;
+
+	tmp = *lst;
+	tmp = tmp->next;
+	new->infile = 1;
+	if (new->fd_in > -1)
```
gwolf-011235, line 35: I think this if clause wants to be below the close() function

src/parser.c
```diff
@@ -0,0 +1,164 @@
...
+{
+	t_cmd	*tmp;
+	t_err	err;
+	int		count_arg;
+	int		count_delim;
+
+	count_arg = 0;
+	count_delim = 0;
+	ft_count_str(lst, &count_arg, &count_delim);
+	err = ft_create_str_arr(&tmp, count_arg, count_delim);
```
gwolf-011235: I would leave the malloc() of the t_cmd itself out of the ft_create_str_arr(), because it hides the mallocing a bit.
But it seems like a line limit problem. Alternative would be to refactor the initilisation step itself (set stuff to null) into a ft_init_cmd() and then just call the function.

src/parser_utils.c
```diff
@@ -0,0 +1,112 @@
...
+#include "mod_parser.h"
+
+/**
+ * @brief Iterate to last node of given list.
+ * 
+ * Returns 0 if list is empty.
+ * @param cmd_head 		List of cmd-structures
+ * @return t_cmd* 
+ */
+t_cmd	*ft_last_cmd(t_cmd *cmd_head)
```
gwolf-011235, line 27: Maybe add "find" to function name: ft_find_last_cmd()

gwolf-011235 (changes requested):
I think we also need a free_cmd funcitons which can be called in case of error. For example if malloc of the delim array fails it should free the arg array and the cmd.
Will be implemented at execution step at the latest, but maybe we should also make an issue to not forget it for this step.

Ima approve without it but it think the position of close() has to be addressed

src/parser_fill_cmd.c
u413-284-si: Why? It is checking whether there already has been an infile opened previously (as new->fd_in is initialised as -1) and in that case it would close it. If not, it will jump over close and proceed to opening the file.

src/parser.c
u413-284-si: It is not (only) a line issue rather than yet again a pointer issue. As I am assigning adresses in ft_create_str_arr() which I'd like to carry back to ft_create_cmd() I need to pass the pointer pointer. That in itself requires me to use a proxy t_cmd *tmp in ft_create_str_arr(). If I don't malloc that proxy, I will get a core dump while compiling. If I am already mallocing, why not just use it for the actual t_cmd that I want to continue using.

src/parser_utils.c
u413-284-si: Thank you for sharing ;D

No, but seriously, don't think that's necessary. It would just make it too verbose and the intention of the function should be clear enough

u413-284-si:
> I think we also need a free_cmd funcitons which can be called in case of error. For example if malloc of the delim array fails it should free the arg array and the cmd. Will be implemented at execution step at the latest, but maybe we should also make an issue to not forget it for this step.
> 
> Ima approve without it but it think the position of close() has to be addressed

All the mallocing should be addressed at the end when we have everything standing and are implementing the cleanup. I think it will be more streamlined to do it in one go than implementing it individually at each step. Will open an issue for it.

src/parser_fill_cmd.c
gwolf-011235: True! I confusion!

src/parser.c
gwolf-011235: Why would you need to malloc a proxy? Since you already have a struct, which let's you pass the pointers directly you don't even need a pointer pointer.

- You can pass the newly allocated t_cmd as t_cmd *tmp
- Then you just save the addresses of the newly malloced arrays in tmp->delim or tmp->args (you can directly set the address. This is because tmp->arg is not a function copy. The "pointer pointer" in this case is *tmp)

src/parser_utils.c
gwolf-011235: Ha! I think we diverge on what is too verbose. But I don't think a change is necessary.

src/parser.c
u413-284-si: These pointer pointers and I will never be friends. Changed, tested, works.

PR #39, opened 2023-07-26
10 builtins

### LINKED ISSUE #10: Builtins
Build some ins

- [ ] cd
- [ ] exit
- [ ] env
- [ ] export
- [ ] unset
- [ ] echo
- [ ] pwd

gwolf-011235:
#### General

- Most of the builtins have their own error handling functions. In case of error (e.g. malloc fail) they print appropriate msg and return t_err. At this point every allocated memory should be free().
- Implemented tests for all of the builtins
- Module builtin depends on Module env for creation of envp array and pwd string.
- ft_exit is not yet finished. I think it's easier after signals and cmd execution have been implemented, to have a concrete idea what should happen.

#### Detail
1.  cd
- Change active directory.
- If no args are given try to change to $HOME.

2. echo
- Print out all args.
- If one (or several) -n flags are given, no newline gets printed. -n flags have to be at beginning and consecutively to be parsed and ignored

3. env
- Print all environment strings. 
- Not sorted (like the "true" env)
- No args are required to be checked according to subject

4. exit
- Call exit() with the given number.
- If more args are given print error
- If no arg is given it should return last exit status - NOT implemented yet. Is a todo in doxygen. Is be easier to implement after signal and command execution is implemented 

5. export
- Insert one or more new environment string.
- If no args are given print the environment alphabetically sorted in "special" form: `export key="value"` (value in quotes). Bash in posix writes export, in non posix it writes declare -x -> used posix format

6. pwd
- Print current working directory
- Doesn't rely on environment value, calls getcwd

7. unset
- Delete one or more environment variables


src/builtin_echo.c
```diff
@@ -0,0 +1,96 @@
...
+ * If at least one -n flag is found, set bool newline to false.
+ *
+ * @param argv NULL terminated args.
+ * @param i Pointer to current index of array.
+ * @param newline Pointer to bool which can get switched.
+ * @return t_err SUCCESS
+ */
+t_err	ft_skip_n_flags(char **argv, size_t *i, bool *newline)
+{
+	while (argv[*i] && !ft_strncmp(argv[*i], "-n", ft_strlen("-n")))
```
u413-284-si: for consistency this should as well not be determined by ft_strlen() rather than a constant number, in this case 3

src/builtin_env.c
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @brief Print all environment variables.
+ *
+ * The strings are not sorted in any way.
+ * Just calls ft_hashtable_print().
+ *
+ * @param env_tab Environment.
+ * @return t_err SUCCESS.
+ */
+t_err	ft_env(t_hashtable *env_tab)
```
u413-284-si, line 29: Function superfluous, you can just call ft_hashtable_print()

src/builtin_exit.c
```diff
@@ -0,0 +1,89 @@
...
+ * @param env_tab Environment.
+ * @return t_err ERR_EXIT
+ * @todo Maybe restructure to not use exit() and handle t_err instead.
+ * @todo If no arg is given should return exit status of last command - global?
+ */
+t_err	ft_exit(char **argv, t_hashtable *env_tab)
+{
+	size_t	size;
+	t_err	err;
+	int		ret_code;
```
u413-284-si, line 39: misaligned variable

src/builtin_unset.c
```diff
@@ -0,0 +1,40 @@
...
+t_err	ft_unset(char **argv, t_hashtable *env_tab)
+{
+	size_t		i;
+
+	i = 1;
+	while (argv[i])
+	{
+		ft_hashtable_delete(env_tab, argv[i], ft_strlen(argv[i]));
+		i++;
+	}
```
u413-284-si, line 37: could save lines with ft_strlen(argv[i++])

src/hashtable_utils.c
```diff
@@ -43,6 +42,7 @@ static size_t	ft_hashtable_index(
  * Check if any param is empty and if key is already inserted.
  * Generate index and malloc new element.
  * Set new element as head of the list pointed to by index.
+ *
```
u413-284-si: no.

src/hashtable_utils.c
```diff
@@ -76,6 +76,7 @@ t_err	ft_hashtable_insert(t_hashtable *ht, char *string, size_t keylen)
  * Check if params are empty.
  * Generate index of given env_var and loop through list.
  * If element is found return the element.
+ *
```
u413-284-si: porque

tests/test_builtin_unset.c
```diff
@@ -0,0 +1,53 @@
...
+	printf(YELLOW"**\tTEST BUILTIN UNSET\t**\n\n"RESET);
+	g_symtab = ft_hashtable_create(1, ft_hash_fnv1);
+	test_setup_argv(5);
+	g_argv[0] = "unset";
+	g_argv[4] = NULL;
+	ft_hashtable_insert(g_symtab, ft_strdup("uno=this"), 3);
+	ft_hashtable_insert(g_symtab, ft_strdup("dos=is"), 3);
+	ft_hashtable_insert(g_symtab, ft_strdup("tr3s=sparta"), 4);
+	ft_hashtable_insert(g_symtab, ft_strdup("quatro=!"), 6);
+	ft_hashtable_insert(g_symtab, ft_strdup("sneaky=boy"), 6);
```
u413-284-si, line 45: boi

u413-284-si:
Minor changes required, nothing substantial.

u413-284-si (changes requested):

src/builtin_echo.c
gwolf-011235: Changed

src/builtin_env.c
gwolf-011235: Chose alternativ way: import hashtable_print() into ft_env. This is also for consitency sake: we have a builtin env and a corresponding function

src/builtin_exit.c
gwolf-011235: Bruh, thats just Github showing lies

src/builtin_unset.c
gwolf-011235: No.

src/hashtable_utils.c
gwolf-011235: Removed all newlines from doxy

src/hashtable_utils.c
gwolf-011235: Removed all newlines from doxy

tests/test_builtin_unset.c
gwolf-011235: He sneaky though

src/builtin_env.c
u413-284-si: So we have two function with almost the same code. Do we still need ft_hashtable_print() then?

src/builtin_exit.c
u413-284-si: schwör alter

src/builtin_unset.c
u413-284-si: 'tis not a valid answer

src/builtin_env.c
gwolf-011235: ft_hashtable_print() has been deleted

src/builtin_exit.c
gwolf-011235: Schwör ich auf yoda!

src/builtin_unset.c
gwolf-011235: For me this change would negatively impact readability for no real upside (at least I don't see one)

- The loop is already doing two things (delete and calc strlen), it would increase complexity
- At first glance - at least for me - it's confusing because the line is handling the same arg, but at same time at the last instance the index gets incremented. You have to use extra brain power to realize it's still the same arg, and incremented for the next run of the loop. 
- The function has plenty of lines left. I think the Norm doesn't want to enforce 'clever' written, hard(er) to understand code with its line limit. It wants to enforce the "do one thing, but do it good" paradigm for functions. Using clever code shortations should be used in rare cases, where the line limit is reached but no other measure can safely be taken. Ask yourself: is the one liner version -exactly- as easy to understand as the five liner? If not: Why shorten it?
- Performance is no issue: the compiler optimizes the code anyway. Even if not: the microseconds in gain are far outweighed by the seconds lost parsing the code (by the human), if any errors are to be investigated.

One could argue that a consistent use of the incrementer in these loops makes the eye more accustomed to it, and therefore increases human parsing speed. But that's kinda like saying we have to write the code in light blue color on dark blue background because with time our comprehension would increase. Why even do it in the first place?

u413-284-si: I think it is not that hard to read as you just follow the execution of the line from left to right and at the end realise that the counter gets incremented after being used in ft_strlen(). As you already pointed out it is a matter of getting accustomed to that kind of code "style". 

Personally I am an advocate of keeping the code as short as possible, especially in cases of if conditions or while loops where the brackets can be saved by having just one line of code within. That facilitates a faster browsing of the code for later maintenance and simply improves coding speed as there is less to type. 

I guess it's a question of keeping the balance between short and sharp code and readability. IMO the intention of the code would rest clear enough with that change, but I leave it to your discretion.

src/builtin_env.c
u413-284-si: Oh, I was referring to ft_hashtable_pretty_print() which resides within hashtable_print.c

gwolf-011235: Pretty print is a debugging function, which also prints additional info like table size and array position of the environmental variables. 
It serves a different purpose, so yes I'd leave it.

src/builtin_unset.c
gwolf-011235: Then I will leave it :D

I think if there would be a bug to trace I would at least second guess myself if I'm really incrementing here correctly. I don't think I can reach any level of accustomisation where the two would be equally parsed by me.

src/builtin_env.c
u413-284-si: So it will not be part of the content we push?

gwolf-011235: You mean relocate it into a testfile?

u413-284-si: no just not put it into the folder that we're gonna upload for eval

gwolf-011235: It was easier to move the function into test_hashtable and delete the file.

PR #41, opened 2023-07-28
40 makefile redesign compiling of tests

### LINKED ISSUE #40: Makefile - redesign compiling of tests
Right now the Makefile deletes main, sets a variable, recompiles. This can be simplified with the definition of two vars for main and testmain

gwolf-011235:
### Redesign of Makefile

- main.c is not in var `SRC` anymore. It's defined in it's own var `OBJ_MAIN`. Same with test_main.c in `TEST_OBJ_MAIN`. This makes compilation between `NAME` and `TEST` easier: no remove of compiled main needed, no debug variables needed
- No src to obj conversion is needed, since these file names don't change. Therefore only an obj variable was defined.
- Since debug var is not needed anymore, removed it out of main.c
- Added new targets `tleak` and `taddress` to compile test with `-fsanitize=leak` or `-fsanitize=address`
- Target `fclean` now also calls `tclean` to remove testfiles
- Added headers to Makefile to make it look chic

Closes #40


Makefile
```diff
@@ -102,19 +147,65 @@ leak: LDFLAGS += -fsanitize=leak
 leak: clean $(NAME)
```
u413-284-si, line 147: Why is the  target leak defined multiple times instead of just once and then have the different steps executed one after the other?

Makefile
```diff
@@ -102,19 +147,65 @@ leak: LDFLAGS += -fsanitize=leak
...
-# TESTING variable which renames "normal" main, and removes the main if it exists.
-# After compiling it removes normal main again, making regular compilation of NAME target possible.
-$(TEST): CFLAGS = -g -DTESTING -gdwarf-4
-$(TEST): prep_test $(TEST_OBJS) $(OBJS)
-	$(CC) $(LDFLAGS) $(TEST_OBJS) $(OBJS) $(LDLIBS) -o $@
-	rm -f obj/main.o
-	@printf "$(GREEN)Starting test!$(RESET)\n"
+# This target adds fsanitize address checker to the flags. It needs to clean and recompile.
+.PHONY: address
+address: CFLAGS += -fsanitize=address
```
u413-284-si, line 152: See above regarding target leak. Same goes for address

Makefile
```diff
@@ -102,19 +147,65 @@ leak: LDFLAGS += -fsanitize=leak
...
+$(TEST): CFLAGS = -g -gdwarf-4
+$(TEST): $(LIBFT) $(OBJS) $(TEST_OBJS) $(TEST_OBJ_MAIN)
+	@printf "\n$(YELLOW)$(BOLD)link test binary$(RESET) [$(BLUE)minishell$(RESET)]\n"
+	$(CC) $(LDFLAGS) $(OBJS) $(TEST_OBJS) $(TEST_OBJ_MAIN) $(LDLIBS) -o $@
+	@printf "\n$(YELLOW)$(BOLD)compilation successful$(RESET) [$(BLUE)minishell$(RESET)]\n"
+	@printf "$(BOLD)$(GREEN)$(TEST) created!$(RESET)\n\n"
+
+# This target adds fsanitize leak checker to the flags. It needs to clean and recompile.
+.PHONY: tleak
+tleak: CFLAGS += -fsanitize=leak
```
u413-284-si, line 182: See target leak

Makefile
```diff
@@ -102,19 +147,65 @@ leak: LDFLAGS += -fsanitize=leak
...
+tleak: LDFLAGS += -fsanitize=leak
+tleak: clean tclean $(TEST)
+	@printf "Compiled with $(YELLOW)$(BOLD)fsanitize=leak$(RESET)\n\n"
+
+# This target adds fsanitize address checker to the flags. It needs to clean and recompile.
+.PHONY: taddress
+taddress: CFLAGS += -fsanitize=address
```
u413-284-si, line 189: See target leak

u413-284-si:
How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c

Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?

Nice formatting! Chic is always asked for

Makefile
gwolf-011235: They are target-specific variable values (see [GNU Make manual 6.11](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html#:~:text=Target%2Dspecific%20variables%20have%20the,variable%20value%20to%20be%20preferred.))

The intent is to (re)define a variable value for a specific target.

For example for `leak` we want to have -fsanitize=leak in CFLAGS and LDFLAGS. We can't (for technical reasons) redefine and have prerequisite on the same line. Therefore a line sets the value for CFLAGS and LDFLAGS respectively. The third line then executes `clean` and `$(NAME)`. (This is also the way how they show it in the manual)
For clarity's sake: Target specific values are also effective for targets called in the recipe. So if I call `$(NAME)` in `leak` CFLAGS will have another value

gwolf-011235:
> How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> 
> Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> 
> Nice formatting! Chic is always asked for

The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. 
The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main).
Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c

For multiple definitions see comment.

Makefile
u413-284-si: Very interesting concept, learned sth new

u413-284-si:
> > How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> > Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> > Nice formatting! Chic is always asked for
> 
> The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main). Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c
> 
> For multiple definitions see comment.

Ok, so you're saying the .c files which are not changing in their form or content in any way, do not need to be listed explicitly in a variable? Is that correct?  But why would main.c not be changing? We are adapting it as much as any other .c file as the program is growing. I am not quite following your reasoning.

gwolf-011235:
> > > How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> > > Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> > > Nice formatting! Chic is always asked for
> > 
> > 
> > The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main). Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c
> > For multiple definitions see comment.
> 
> Ok, so you're saying the .c files which are not changing in their form or content in any way, do not need to be listed explicitly in a variable? Is that correct? But why would main.c not be changing? We are adapting it as much as any other .c file as the program is growing. I am not quite following your reasoning.

No. The size and content of the **SRC** variable is constantly changing. 

We have to add all source files explicitly in the Makefile, no wildcards are allowed.
If we could use a wildcard, we still would need to change this list of .c files into a list of .o files which can be used as a prerequisite list.
With the project layout (subdirectory src, obj, ..) we also need to append some different prefixes, to let `make` know where to search the files.
Right now we just add a .c file to the SRC variable. Then prefixes and suffixes are changed in the variables SRCS and OBJS.

Alternatively we could just have a OBJ variable, where we don't list the .c files but just the (to be created) .o files. Then we would just need to add the prefix OBJ_DIR. 
This would still work since the target for compiling an OBJ_DIR/%.o is searching for a complementing file in the SRC_DIR/%.o. At this stage the content of the SRC variable is meaningless. 

We are doing it this shorter way with the main files. Its even shorter since we already added the prefix OBJ_DIR. The content of these files may change but the name stays the same. So for this special files we have them explicitly as OBJ_DIR/.o files.

We could also define two extra SRC_MAIN variables, another two for adding prefix, another two for substituting suffix, another two for adding other prefix. 
But in the end reach the same goal as with just two variables.

As I've said I don't think it's Norm conform. But we can just add the main.c to SRC and delete the tests and are golden.

u413-284-si:
> > > > How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> > > > Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> > > > Nice formatting! Chic is always asked for
> > > 
> > > 
> > > The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main). Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c
> > > For multiple definitions see comment.
> > 
> > 
> > Ok, so you're saying the .c files which are not changing in their form or content in any way, do not need to be listed explicitly in a variable? Is that correct? But why would main.c not be changing? We are adapting it as much as any other .c file as the program is growing. I am not quite following your reasoning.
> 
> No. The size and content of the **SRC** variable is constantly changing.
> 
> We have to add all source files explicitly in the Makefile, no wildcards are allowed. If we could use a wildcard, we still would need to change this list of .c files into a list of .o files which can be used as a prerequisite list. With the project layout (subdirectory src, obj, ..) we also need to append some different prefixes, to let `make` know where to search the files. Right now we just add a .c file to the SRC variable. Then prefixes and suffixes are changed in the variables SRCS and OBJS.
> 
> Alternatively we could just have a OBJ variable, where we don't list the .c files but just the (to be created) .o files. Then we would just need to add the prefix OBJ_DIR. This would still work since the target for compiling an OBJ_DIR/%.o is searching for a complementing file in the SRC_DIR/%.o. At this stage the content of the SRC variable is meaningless.
> 
> We are doing it this shorter way with the main files. Its even shorter since we already added the prefix OBJ_DIR. The content of these files may change but the name stays the same. So for this special files we have them explicitly as OBJ_DIR/.o files.
> 
> We could also define two extra SRC_MAIN variables, another two for adding prefix, another two for substituting suffix, another two for adding other prefix. But in the end reach the same goal as with just two variables.
> 
> As I've said I don't think it's Norm conform. But we can just add the main.c to SRC and delete the tests and are golden.

Basically what I am taking from this is that the listing of the source files as such is completely unnecessary. We could just name all the objects to be created and be done with it. The detour with the src files is only due to the norm which is demanding us to name all source files explicitly which are required to compile the program.

The separation of the main.c from all other .c files (= OBJS) on the other hand allows us to compile the test_main.c with OBJS without having a conflict of two mains. And you chose to compile both mains with this shortcut approach which is not norm-conform.

Does that summarise it?

Well, if we are to leave it like that, I'd strongly suggest to open an issue to adapt the Makefile before pushing. Otherwise we risk to forget it.

gwolf-011235:
> > > > > How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> > > > > Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> > > > > Nice formatting! Chic is always asked for
> > > > 
> > > > 
> > > > The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main). Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c
> > > > For multiple definitions see comment.
> > > 
> > > 
> > > Ok, so you're saying the .c files which are not changing in their form or content in any way, do not need to be listed explicitly in a variable? Is that correct? But why would main.c not be changing? We are adapting it as much as any other .c file as the program is growing. I am not quite following your reasoning.
> > 
> > 
> > No. The size and content of the **SRC** variable is constantly changing.
> > We have to add all source files explicitly in the Makefile, no wildcards are allowed. If we could use a wildcard, we still would need to change this list of .c files into a list of .o files which can be used as a prerequisite list. With the project layout (subdirectory src, obj, ..) we also need to append some different prefixes, to let `make` know where to search the files. Right now we just add a .c file to the SRC variable. Then prefixes and suffixes are changed in the variables SRCS and OBJS.
> > Alternatively we could just have a OBJ variable, where we don't list the .c files but just the (to be created) .o files. Then we would just need to add the prefix OBJ_DIR. This would still work since the target for compiling an OBJ_DIR/%.o is searching for a complementing file in the SRC_DIR/%.o. At this stage the content of the SRC variable is meaningless.
> > We are doing it this shorter way with the main files. Its even shorter since we already added the prefix OBJ_DIR. The content of these files may change but the name stays the same. So for this special files we have them explicitly as OBJ_DIR/.o files.
> > We could also define two extra SRC_MAIN variables, another two for adding prefix, another two for substituting suffix, another two for adding other prefix. But in the end reach the same goal as with just two variables.
> > As I've said I don't think it's Norm conform. But we can just add the main.c to SRC and delete the tests and are golden.
> 
> Basically what I am taking from this is that the listing of the source files as such is completely unnecessary. We could just name all the objects to be created and be done with it. The detour with the src files is only due to the norm which is demanding us to name all source files explicitly which are required to compile the program.
> 
> The separation of the main.c from all other .c files (= OBJS) on the other hand allows us to compile the test_main.c with OBJS without having a conflict of two mains. And you chose to compile both mains with this shortcut approach which is not norm-conform.
> 
> Does that summarise it?
> 
> Well, if we are to leave it like that, I'd strongly suggest to open an issue to adapt the Makefile before pushing. Otherwise we risk to forget it.

Yes good summary.

Will create a issue

u413-284-si:
> > > > > > How come that main.c does not need to be mentioned explicitly in the Makefile as a variable which is to be converted into a main.o in the variable OBJ_MAIN? Same applies to test_main.c
> > > > > > Furthermore it is curious that for the targets leak, address, tleak and taddress are defined each multiple times. Is that intentionally or just a hiccup?
> > > > > > Nice formatting! Chic is always asked for
> > > > > 
> > > > > 
> > > > > The reason for doing it with SRC and OBJ is not to compile them into objects: we just want a list of all objects. We get it by defining all source files. Then do some Makefile magic to append the SRC_DIR, also replace all .c with .o and append the OBJ_DIR. Since the main isn't gonna change any way it can just write it out without having it in it's own SRC_MAIN variable. The main.o is a dependency. The rule for compiling .c to .o files is searching in the SRC_DIR, where it's gonna find the main.c (same goes for test_main). Hardcore norm guys could argue that since the main.c is not explicitely listed, it's an error. For pushing I would generelly leave out the tests directory and rewrite the SRC to include main.c
> > > > > For multiple definitions see comment.
> > > > 
> > > > 
> > > > Ok, so you're saying the .c files which are not changing in their form or content in any way, do not need to be listed explicitly in a variable? Is that correct? But why would main.c not be changing? We are adapting it as much as any other .c file as the program is growing. I am not quite following your reasoning.
> > > 
> > > 
> > > No. The size and content of the **SRC** variable is constantly changing.
> > > We have to add all source files explicitly in the Makefile, no wildcards are allowed. If we could use a wildcard, we still would need to change this list of .c files into a list of .o files which can be used as a prerequisite list. With the project layout (subdirectory src, obj, ..) we also need to append some different prefixes, to let `make` know where to search the files. Right now we just add a .c file to the SRC variable. Then prefixes and suffixes are changed in the variables SRCS and OBJS.
> > > Alternatively we could just have a OBJ variable, where we don't list the .c files but just the (to be created) .o files. Then we would just need to add the prefix OBJ_DIR. This would still work since the target for compiling an OBJ_DIR/%.o is searching for a complementing file in the SRC_DIR/%.o. At this stage the content of the SRC variable is meaningless.
> > > We are doing it this shorter way with the main files. Its even shorter since we already added the prefix OBJ_DIR. The content of these files may change but the name stays the same. So for this special files we have them explicitly as OBJ_DIR/.o files.
> > > We could also define two extra SRC_MAIN variables, another two for adding prefix, another two for substituting suffix, another two for adding other prefix. But in the end reach the same goal as with just two variables.
> > > As I've said I don't think it's Norm conform. But we can just add the main.c to SRC and delete the tests and are golden.
> > 
> > 
> > Basically what I am taking from this is that the listing of the source files as such is completely unnecessary. We could just name all the objects to be created and be done with it. The detour with the src files is only due to the norm which is demanding us to name all source files explicitly which are required to compile the program.
> > The separation of the main.c from all other .c files (= OBJS) on the other hand allows us to compile the test_main.c with OBJS without having a conflict of two mains. And you chose to compile both mains with this shortcut approach which is not norm-conform.
> > Does that summarise it?
> > Well, if we are to leave it like that, I'd strongly suggest to open an issue to adapt the Makefile before pushing. Otherwise we risk to forget it.
> 
> Yes good summary.
> 
> Will create a issue

Alright, than let's bring that baby on board! Seems that there are some minor conflicts (mostly comments or header conflicts). As you're working tmr, I will just do it first thing in the morning

PR #48, opened 2023-08-08
8 executor

### LINKED ISSUE #8: Executor
Use cmd tables from parser and execute them

- [ ] Simple cmds (echo hallo)
- [ ] Pipelines (env | grep boy)
- [ ] Check where is the command

### LINKED ISSUE #45: executor issues
1. waitpid at end of whole pipe needed ?
2. heredoc with indices
3. unknown cmd in pipeline -> what happens with the associated pipe (not opened or close write end?)

### LINKED ISSUE #47: Handle built-ins in executor
Requires:
- check for built-ins
- execution of builit-ins (incl. call of correct built-in function)
- differentiation: single cmd vs. piped built-ins
- forking of certain built-ins

u413-284-si:
Executes given cmds in form of a linked list t_cmd *cmd:

Single cmd:
-> Built-in: executed directly by function call
-> system calls: forked into child process and executed within via execve

Pipes:
-> Built-in: forked into child process and executed within via function call
-> system calls: forked into child process and executed within via execve
-> piping of cmd output into next cmd input
-> precedence of individual infiles, outfiles, heredocs over pipe: fd are set accordingly

- all forked processes are waited for by parent after all child processes have been created
- heredocs are handled: several heredocs are opened, take input and are unlinked if more follow. last heredoc remains open as input to cmd if no infile follows. If infile follows, all heredocs are unlinked and infile set as input.
- unknown cmds: error message is given, rest of pipe is still executed

closes #8
closes #45
closes #47


inc/mod_handle_input.h
```diff
@@ -40,7 +42,7 @@ typedef enum e_type
 	OUTFILE,
 	APPEND,
 	PIPE,
-	NEWLINE
+	NWL
```
gwolf-011235: Since none of the other one are abbreviated I suggest NEW_LINE (with underscore to differentiate to NEWLINE of readline)
Or if unsystematic: NLINE is a bit clearer

src/cleanup.c
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @file cleanup.c
+ * @brief General functions to clean up the program.
+ */
+
+#include "mod_cleanup.h"
+
+/**
+ * @brief Frees a string array and its strings.
```
gwolf-011235: I would clarify:
"Frees a NULL-terminated string array and its strings."

src/executor.c
```diff
@@ -0,0 +1,143 @@
...
+{
+	t_err	err;
+	char	**paths;
+
+	paths = NULL;
+	ft_init_exec(cmd);
+	err = ft_handle_heredoc(cmd);
+	if (err != SUCCESS)
+		return (err);
+	err = ft_create_pipes(cmd);
```
gwolf-011235, line 49: Short comment in doxy what it does pls

src/executor.c
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * @brief Executes list of cmds provided.
+ * 
+ * Create pipes inbetween cmds.
+ * 		fd_pipe[0] refers to the read end of the pipe.
+ * 		fd_pipe[1] refers to the write end of the pipe.
+ * Data written to the write end of the pipe is buffered 
+ * by the kernel until it is read from the read end of the pipe.
```
gwolf-011235: Move up

src/executor_builtin.c
```diff
@@ -0,0 +1,88 @@
...
+ * @file executor_builtin.c
+ * @brief Functions to handle execution of built-ins.
+ */
+
+#include "mod_executor.h"
+
+/**
+ * @brief Verify if given cmd is a builtin.
+ * 
+ * Return true if found, else false.
```
gwolf-011235: The Line regarding the return should be split and written at the bottom ( @return true and @return false)

src/executor_cleanup.c
```diff
@@ -0,0 +1,169 @@
...
+
+#include "mod_executor.h"
+
+void	ft_cleanup_cmd(t_cmd *cmd)
+{
+	ft_close(cmd->fd_in);
+	ft_close(cmd->fd_out);
+	ft_plug_pipe(cmd, 1);
+	if (cmd->heredoc)
+		free(cmd->heredoc);
```
gwolf-011235: Why not ft_unlink_heredoc()?

src/executor_cleanup.c
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * @brief Unlinks the temporary heredoc file created.
+ * 
+ * Returns provided error code.
+ * @param name 		Name of heredoc to be unlinked.
+ * @param err 		Provided error code.
+ * @return t_err 	err
+ */
+t_err	ft_unlink_heredoc(char *name, t_err err)
```
gwolf-011235: If char **name gets passed, this function could also set the pointer to NULL

src/executor_cleanup.c
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * @brief Closes pipe ends (= file descriptors) of the
+ * specified cmd.
+ * 
+ * fd_pipe[0] refers to the read end of the pipe.
+ * fd_pipe[1] refers to the write end of the pipe.
+ * @param cmd		Current cmd.
+ * @return t_err	ERR_CLOSE, SUCCESS
+ */
+t_err	ft_plug_pipe(t_cmd *cmd, bool close_read_end)
```
gwolf-011235: This function closes write end, and if bool is true the read end. 
I think ft_close(pipe[1]) does the same thing as bool set to false? Can we get rid of bool, so it always closes both and use ft_close if just wirte end needs to be closed?

It also seems that there is only one instance - in cleanup - where this bool is set to true.

src/executor_create_child.c
```diff
@@ -0,0 +1,140 @@
...
+	else
+		err = ft_raise_middle(cmd, envp, data, builtin);
+	return (err);
+}
+
+/**
+ * @brief Creates a child by forking and calls the pertinent
+ * function which is to be executed by the child.
+ * 
+ * The parent does not wait for the child to finish (flag WNOHANG set), it
```
gwolf-011235: Is old comment from pipex, also below

src/executor_heredoc.c
```diff
@@ -0,0 +1,153 @@
...
+	name = NULL;
+	err = ft_initiate_heredoc(cmd->index, &name, &fd);
+	if (err != SUCCESS)
+		return (err);
+	len = ft_strlen(delim);
+	while (1)
+	{
+		write(1, "pipe heredoc> ", 14);
+		buf = get_next_line(0);
+		if (!buf)
```
gwolf-011235, line 72: I think gnl returns NULL if EOF is sent (Strg+D). This is the case were bash writes a warning, but still processes the heredoc.

Since this function is already packed I think we need to refactor the reading from gnl into its own function (we also need the check for g_status), I suggest a comment to remember refactor

src/executor_heredoc.c
```diff
@@ -0,0 +1,153 @@
...
+{
+	int		i;
+	t_err	err;
+
+	while (cmd)
+	{
+		i = -1;
+		while (++i < cmd->delim_pos)
+		{
+			err = ft_create_heredoc(cmd, cmd->delims[i], i);
```
gwolf-011235, line 36: I think you need to pass cmd->index as curr_index. 
If cmd-1 and cmd-2 have one heredoc, they would have the same name. Or am I missing something?

src/executor_heredoc.c
```diff
@@ -0,0 +1,153 @@
...
+ * @param cmd 		Current cmd being processed.
+ * @param name 		Name to be defined.
+ * @return t_err 	ERR_MALLOC, SUCCESS
+ */
+t_err	ft_name_heredoc(int index, char **name)
+{
+	char	*index_str;
+
+	index_str = ft_itoa(index);
+	*name = ft_strjoin(".tmp_heredoc_v", index_str);
```
gwolf-011235, line 97: Itoa mallocs, needs to be freed in case of error in strjoin

src/executor_create_child.c
```diff
@@ -0,0 +1,140 @@
...
+		if (err != SUCCESS)
+			return (err);
+		return (ERR_FORK);
+	}
+	else if (cmd->pid == 0)
+		ft_firstborn(cmd, envp, data, builtin);
+	err = ft_plug_pipe(cmd, 0);
+	if (err != SUCCESS)
+		return (err);
+	cmd->fd_pipe[1] = -1;
```
gwolf-011235: Why do we set pipe[1] to -1 here?
I remember you said something about setting them to -1, but I thought you said that ft_close sets them to -1, but it's "just" checking if they are bigger than -1. Could ft_close also set them to -1 or would this break something?

src/executor_child.c
```diff
@@ -0,0 +1,149 @@
...
+		ft_close(cmd->fd_out);
+	}
+	else
+	{
+		if (cmd->fd_in >= 0)
+			ft_replace_fd(cmd->fd_in, cmd->fd_pipe[1]);
+		else
+			ft_replace_fd(0, cmd->fd_pipe[1]);
+		ft_close(cmd->fd_pipe[1]);
+	}
```
gwolf-011235, line 52: This part - dup2 of fd_out and fd_in - repeats in all three -born functions. 
It could be refactored to make error handling possible. Right now we generally ignore all syscall fails (the functions throw an error, but is never retrieved). 
I'm not sure if we should/can continue if dup2 fails, but with two smaller functions (ft_set_fdout and ft_set_fdin or so), the higher function could at least pass it on.

Can also be looked at while checking the whole error handling, but maybe make a comment for easy refactor?

src/executor_utils.c
```diff
@@ -0,0 +1,155 @@
...
+ */
+t_err	ft_check_cmd_access(char **args, char **cmd_paths)
+{
+	t_err	err;
+
+	err = ERR_UNKNOWN_CMD;
+	if ((*args)[0] == '/' || !ft_strncmp(*args, "./", 2)
+		|| !ft_strncmp(*args, "../", 3) || !cmd_paths)
+	{
+		if (access(*args, F_OK | X_OK) == 0)
```
gwolf-011235, line 46: I think we need a wrapper around access to catch an error and print corresponding error message

src/executor_utils.c
```diff
@@ -0,0 +1,155 @@
...
+
+	while (*envp && (ft_strncmp(*envp, "PATH", 4) || (*(*envp + 4) != '=')))
+		envp++;
+	if (!*envp)
+		return (ERR_NOPATH);
+	path_str = *envp + 5;
+	*paths = ft_split(path_str, ':');
+	if (!*paths)
+		return (ERR_MALLOC);
+	if (***paths == '\0')
```
gwolf-011235: Random thought: can ft_split generate empty strings? For example PATH="::/usr/bin:/usr/sbin" (two colon at the start). If yes this if would kill PATH

src/executor_utils.c
```diff
@@ -0,0 +1,155 @@
...
+	return (SUCCESS);
+}
+
+/**
+ * @brief Wait for child processes
+ * 
+ * @param cmd 		List of cmds.
+ * @return t_err 	ERR_WAIT, SUCCESS
+ */
+t_err	ft_wait_for_babies(t_cmd *cmd)
```
gwolf-011235, line 148: This function also needs to decipher the return value. I suggest a comment so we don't forget

src/main.c
```diff
@@ -29,7 +30,8 @@ int	main(int argc, char **argv)
 		ft_exit_failure(data, ERR_MALLOC);
 	if (ft_env_setup(&data->env_table) != SUCCESS)
 		printf("NO\n");
-	ft_hashtable_insert(data->env_table, "PS1=\\u@\\h:\\w$ ", 3);
+	ft_envp_create(data->env_table, &envp);
```
gwolf-011235: This has to happen in the while loop, so it can be updated. But main needs a touchup anyway

src/parser_utils.c
```diff
@@ -104,3 +104,27 @@ t_err	ft_create_str_arr(t_cmd *tmp, int count_arg, int count_delim)
 		tmp->delims = NULL;
 	return (SUCCESS);
 }
+
+/**
+ * @brief Initialise cmd values.
+ * 
+ * @param tmp Cmd being initialised.
+ */
+void	ft_init_cmd(t_cmd *tmp)
```
gwolf-011235, line 113: I think the init of the values should follow the order of the struct.

Also: when using bool I would  assign false/true and not 0/1

tests/test_executor.c
gwolf-011235, line 1: For tests in general: I think it's a good habit to setup g_data or envp or anything, what can be reused, in the driver function of the test. The wrapper function, which (ideally only) calls the tested function can then use the initialized g_vars.
At the end of the test a cleanup of the setup variables makes it possible to test memory leaks, since ideally all data created by the test would be free'd.

I also think that this file could be split into several since its already so big. For example a testfile for heredocs seems sensible.

gwolf-011235 (commented):
Looks good.

One - easy fix - memory leak when creating heredocs. 

Error handling right now is all over the place - but this is a general theme in the repo right now. As you said: it's best if we look into it together.

Some refactor possibilities, which are not priority but could be remembered in a comment.

Doxygen comments in this module are a bit lacking, and don't live up to the usual standard. Sometimes copied from other repo, often times repeating (I think we don't need to repeat everywhere that pipe[1] is the write side). There is also a file with huge parts commented out (I think if we want it for later we could just reference your pipex again).
I only commented on a few of them.
Understandable since time constraints but when we have a tested prototype we should look into them. But very industry standard - nobody updates the comments :D

inc/mod_handle_input.h
u413-284-si: Argument is abbreviated to ARG as the full name would be too long. To stay in accordance with that I suggest NEWL, if you rly are against the beloved NWL

src/cleanup.c
u413-284-si: It seems that it isn't forcibly a standard in C to NULL-terminate a string, even less so a string array. Therefore agreed to specify this standard for the handling of string arrays in our project. Comment adapted.

src/executor.c
u413-284-si: Done.

src/executor.c
u413-284-si: Transferred to doxygen of ft_create_pipes()

src/executor_builtin.c
u413-284-si: Both return values are at the end of the doxygen. Removed the description stating the obvious.

src/executor_cleanup.c
u413-284-si: Function not finished yet. Will be used for the final cleanup which is to be implemented yet. Added err checks, ft_unlink_heredoc() and doxygen. To be completed later on.

src/executor_cleanup.c
u413-284-si: After freeing name? Isnt that the case already once it is freed?

src/executor_cleanup.c
u413-284-si: Reasonable enough. Changed as suggested.

src/executor_create_child.c
u413-284-si: updated.

src/executor_heredoc.c
u413-284-si: Comment added and issue created.

src/executor_heredoc.c
u413-284-si: The heredoc name is created with cmd->index. See line 64:

err = ft_initiate_heredoc(cmd->index, &name, &fd);

The counter variable i serves to know whether the last heredoc has been reached (in case of several heredocs in a cmd). In that case the heredoc is not unlinked, but retained and its fd assigned to cmd->fd_in

src/executor_heredoc.c
u413-284-si: Needs to be freed in any case, if error or success of ft_strjoin(). Added frees for both cases.

src/executor_create_child.c
u413-284-si: Had the same idea this evening while in the gym. Already changed that, ft_close sets fd to -1 now. Explicit instances removed.

u413-284-si: And no, nothing broke, all tests are still viable.

src/executor_child.c
u413-284-si: The devil lies in the detail. Every function duplicates different fds.

The errors are not retrieved on purpose as they are created in the child process and I don't see how they can be returned to the parent. The parent process will only get the exit status via pid, any other information of the child will be lost.

src/executor_utils.c
u413-284-si: Not sure we need to catch that "error" because when it's returning -1, it basically is just saying that it didn't find the right path+cmd combination yet. That's why it's looping through all paths. When rly no good combo is found, then the according error message is printed by ft_process_cmd()

src/executor_utils.c
u413-284-si: In theory yes, ft_split would have an empty string with '\0'. But that should never occur as PATH is constant and is never with an empty value between colons. And I do not think that the shell is supposed to be able to handle the case when someone willfully unsets and resets PATH to this form.

src/executor_utils.c
u413-284-si: You mean the exit status of the child process? Comment set.

src/main.c
u413-284-si: Yes, main definitely is not in any way near its final form and should be ignored until we actually start bringing it up to speed.

src/parser_utils.c
u413-284-si: I can understand the need to clarify the bools with false/true statements, but the order... Rly is not of relevance :D

tests/test_executor.c
u413-284-si: The tests do not follow any of our coding guidelines and are rather sloppily generated, just enough to fulfill their goal to verify a certain functionality. Therefore I would not spent to much time and effort on making them presentable. When we move on to the cleanup and valgrinding, we will take care of the necessary freeing in the tests in order to differentiate between test and program variables.

u413-284-si:
> Looks good.
> 
> One - easy fix - memory leak when creating heredocs.
> 
> Error handling right now is all over the place - but this is a general theme in the repo right now. As you said: it's best if we look into it together.
> 
> Some refactor possibilities, which are not priority but could be remembered in a comment.
> 
> Doxygen comments in this module are a bit lacking, and don't live up to the usual standard. Sometimes copied from other repo, often times repeating (I think we don't need to repeat everywhere that pipe[1] is the write side). There is also a file with huge parts commented out (I think if we want it for later we could just reference your pipex again). I only commented on a few of them. Understandable since time constraints but when we have a tested prototype we should look into them. But very industry standard - nobody updates the comments :D

The mem leak has been adressed. 
Doxygen documentation has been amended, but a final sweep at the end is anyways necessary as we will certainly do some changes while finishing up the program (including the last functionalities and during testing). When we are content with its functionality and all mem leaks/fds are taken care of, we need to:
- update the documentation
- check the formatting
- remove unwanted comments 
- do a final round of testing after everything

inc/mod_handle_input.h
gwolf-011235: I accept NEWL

src/executor_builtin.c
gwolf-011235: Github doesn't show the new version interestingly, so I can't see the current state. 
You could add at to the returns just a little bit i.e. @true Found and @false Not found. Just so they don't dangle empty

src/executor_utils.c
gwolf-011235: PATH is not constant. You can add and remove stuff as you like, that's the purpose of PATH. 

Examples: 

- If you want to exe your little progs in your home directory you can add the home directory to PATH. Then just type the program name.
- If you want to little what users can just exe, you can remove some directories from PATH.
- If you want to trip someone during eval you can unset PATH and see what happens.

Since PATH is can be altered as you like it can also be 'wrong'. Question is what do we have to catch. Bash doesn't care if there are empty colons at the start.

src/executor_cleanup.c
gwolf-011235: free() only frees the malloc'd space associated with the value (address) of the passed pointer. Technically free() cannot change the value of the pointer, since you just pass the value and not the reference to the pointer itself.
That's also the reason why we would need to pass a char** to set it to NULL.

If we never look at the pointer again it doesn't make a difference, but since we already have a dedicated cleanup function, we should also set the pointer to NULL.

src/executor_cleanup.c
gwolf-011235: Just to clarify: did you swap all instances of plug_pipe to ft_close, where the bool was set to false?

src/executor_create_child.c
gwolf-011235: Brain gym :O

src/executor_cleanup.c
u413-284-si: Yes, and ft_plug_pipes() doesn't have a bool anymore.

src/executor_heredoc.c
gwolf-011235: Sloppy reading on my part, thanks for clarifying!

src/executor_utils.c
gwolf-011235: Depends on how "deep" we want to go into error handling: access (or nearly all syscalls) set different errno values depending on the encountered error. We could wrap to catch errno and print the corresponding errno message with perror(). 
I'll add a "maybe" issue

src/executor_child.c
gwolf-011235: Even if they dupl different: you can pass different vars, depending on what -born you are. 
Since it's more of a refactor story, we can leave it as is.

Except on question: ft_lastborn closes fd_out in any case, outside of the if condition. first and middle only if fd_out is >= 0 (which seems more logical to me). Is this an oversight?

The errors can be retrieved via the exit status of the child. Even if it's just a 1 instead of 0 it would indicate an error.

src/executor_utils.c
gwolf-011235: Fun fact: adding a : in front or at the end of $PATH allows you to execute every executable in the same directory (exe instead of typing ./exe). 
So we don't even need to clean out empty strings: "" + cmd = cmd so this should execute?

inc/mod_handle_input.h
u413-284-si: Refactored in all instances

src/executor_builtin.c
u413-284-si: I'd say that this is your opportunity to show your trust in me... 

But anyways doxygen is updated.

src/executor_utils.c
u413-284-si: Now that is an interesting behaviour. In that case an empty string in paths will lead to paths being freed and its pointer set to NULL. When checking for the cmd accessibility that case will be caught by

if ((*args)[0] == '/' || !ft_strncmp(*args, "./", 2)  || !ft_strncmp(*args, "../", 3) || !cmd_paths)

ergo the last condition. That covers the behaviour of each cmd being directly executable in its own directory. If it is not in its directory, an error message type of "cmd not found" will be send. 

So basically the status quo is good.

src/executor_cleanup.c
u413-284-si: Very good, ran some tests and can confirm that simply freeing the string sets its address to some garbage, not NULL. 

Technically it does not matter whether we set it to NULL though as it will either be overwritten by a new heredoc name or never touched again. But out of principle to have everything neatly reset, just as we set closed fds to -1, the desired performance is now implemented.

src/executor_child.c
u413-284-si: The closing of fd_out may be moved inside the if condition fd_out >= 0. Technically it doesn't matter as ft_close() contains a check whether the fd rly needs to be closed and doesn't if fd was not set. But just for the beauty of uniformity it is now inside the if cond.

src/executor_builtin.c
gwolf-011235: Pull requests are not a place for trust only hard facts :D

src/executor_utils.c
gwolf-011235: No that's still a problem: we can _only_ execute in the same directory. If you add an empty string to PATH it gives you the possibility to _also_ execute in current directory.
If exe is not there, check in other PATHs

We currently behave differently if empty string is 
at start (clear PATH) or 
in the middle or at the end (leave PATH including empty string)

I think the solution is simple: just not check if there is an empty string. If ft_split is success it returns an array (which is checked). If we have an array we don't care what the strings are. Could be twenty empty strings and /usr/lib - we just use all of them.

u413-284-si: Let me get this straight: 

If there is an empty string in PATH -> 
1) we can execute the cmd, if we're currently in the same directory (no need for ./)
2) if we are not in the same directory, it will iterate through PATH and try to find the right directory: if its directory is not in PATH, it will display an error. If its directory is in PATH, it will execute

Is that the behaviour you're describing? If yes, that would mean that the code goes into both cases of ft_check_cmd_access():

	if ((*args)[0] == '/' || !ft_strncmp(*args, "./", 2)
		|| !ft_strncmp(*args, "../", 3) || !cmd_paths)
	{
		if (access(*args, F_OK | X_OK) == 0)
			return (SUCCESS);
	}
	else
		err = ft_prefix_path(args, cmd_paths);

Do both, first the access check with the cmd in the current directory and if that's not successfull, try to prefix the right path.

gwolf-011235: Not exactly. The behavior is just "iterate through PATH". Let's say PATH is /usr/bin/:/usr/sbin/: which should split into
1. "/usr/bin/"
2. "/usr/sbin/"
3. ""

After strjoin with cmd we check access of
1. /usr/bin/cmd
2. /usr/sbin/cmd
3. cmd

If we have no path we would also check the same as 3, but that's just because we can't append anything.

u413-284-si: But isn't case 3 just try to execute the cmd as is in the current directory?

If directory is the directory of the executable, then success. If not, then error.

gwolf-011235: Yes it does have the same effect but we don't even need to check anything extra.

Why do a directory check if everything works if we just delete the if condition?

What is the purpose of the if condition in lines 109-114?

u413-284-si: I think we do. Cause without it in case of an empty string, we would go into the second case and try to prefix the cmd. This would result in /cmd and not cmd. Then when checking the access an error will occur and we will not be able to execute the cmd even if we're in the same directory.

So in order to cover the desired behaviour we will need to go into both cases. But we need to adapt the if condition as it will delete our whole PATH currently which is not what we want. Instead it shall set a flag that allows the cmd to enter the first case and leave the PATH as is (or remove the empty string as it will serve nothing).

gwolf-011235: You would need to check the whole array since an empty string can be anywhere and more than once.

I.e. :::/usr/lib/:::

Why should this happen? Because eval.

Alternative: if empty string: don't add '/'

u413-284-si: Had a closer look at ft_split(): It doesn't create empty strings in case of ::, it just jumps over these instances.

In order to catch this we need to write an extra function that sets a flag in case it encounters double colons in the PATH. Only then we can handle this special behaviour.

gwolf-011235: That sucks :( 

Other special cases:

- PATH=: (just colon also works)
- PATH=/usr/lib/: (adding just one colon at the end also works) 

How does ft_split handle PATH="" (just empty string)?

u413-284-si: Need to check these cases individually to be sure, but I think:
PATH=: creates an array with just NULL inside
PATH=/usr/lib/: creates an array with /usr/lib/ and NULL
PATH="" creates an array with just NULL inside

What is the effect of these cases on bash behaviour that we need to imitate?

gwolf-011235: All three make it possible to execute exe in current directory. 

I think you have to see it as empty string equals exe in current directory (since cmd + "" = cmd):

PATH=: equals two PATHs (left and right side of colon). Both are empty string.
PATH=/usr/lib/: equals two PATHs. Right side of colon is empty string.
PATH="" equals one PATH. Empty string.

u413-284-si: Confirmed my assumptions in previous comment. ft_split() behaves in those 3 cases as predicted.

Created a ft_check_empty_path() function that sets a flag in all 3 cases and when :: indicating that an empty path was detected. Said flag needs to be passed on to ft_check_cmd_access() residing inside of ft_execute_scmd() and ft_execute_pcmd(). This requires char **paths to be integrated into t_data *data as all 4 input parameters are already taken in ft_execute_scmd() and ft_execute_pcmd().

gwolf-011235: Conferenced - looks good!

PR #54, opened 2023-08-13
37 env variables dont necessarily need a value

### LINKED ISSUE #37: Env variables don't necessarily need a value
Export allows variables, which have no value. The snippet
`export hello`
exports the variable hello, with no value. 

Currently the environment uses the same string for name and value, just different pointers. Value points to the char after the first equals sign. This could lead to some problems which need to be investigated:

- I think the value should point to NULL if no equals is found. But since it points to the zero terminator if no euqals is found, it could also work.
- I think the equals is also used for getting the variable name. This will have to be adapted.
- An exported variable with no value won't be in the child's environment. This needs to be checked when creating the envp.
- If everything falls apart a rewrite could split the name and value. At first glance the same problems would persists (check if no value, check how long variable name)

gwolf-011235:
This branch solves an invalid read, with potential of SIGV, when an empty variable was added to the environment.

1. Update hashtable

`export a` exports a variable, but the variable has no value. This variable doesn't show up in the list printed by env, but does show up in the (non standard) list of export in bash --posix. Minishell now behaves the same.

- environment variables have a new bool `has_value` to check if they have a value. Depending on that bool the value pointer is set either after the equals sign or after the variable name. The has_value check has to be done before insertion. This bool needs to be passed to `hashtable_insert()` and `hashtable_swap()`.
- The hashtable counts the number of elements in the table with `num_elements`. Now it also counts the variables with value assigned with `num_values`.
- `ft_envp_create()` only adds variables which have a value into the envp. The new `ft_envp_create_all()` adds all variables to the array.
- `ft_env()` uses the first, `ft_export()` uses the latter function.

2. Refactor

Refactored tests to fit the new function prototypes of  `hashtable_insert()` and `hashtable_swap()`.

Closes #37

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,8 @@
 		"string_view": "c",
 		"mod_expand.h": "c",
 		"minishell_config.h": "c",
-		"mod_prompt.h": "c"
+		"mod_prompt.h": "c",
+		"mod_env.h": "c",
+		"mod_builtin.h": "c"
 	}
 }
--- a/inc/hashtable.h
+++ b/inc/hashtable.h
@@ -6,7 +6,7 @@
 /*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/05/15 18:07:23 by gwolf             #+#    #+#             */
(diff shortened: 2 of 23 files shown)


u413-284-si:
No errors detected, field testing of functionality follows.

PR #57, opened 2023-08-14
53 str+c in heredoc

### LINKED ISSUE #49: Heredoc behaviour: SIGNAL Str+D
In ft_create_heredoc():

Does get_next_line() need to be adapted to exhibit the correct behaviour for Str+D while reading input for the heredoc?
-> requires further research and possible refactor

### LINKED ISSUE #53: STR+C in heredoc
Cancels whole pipe 
-> previously created outfile need to be closed and unlinked
-> no cmds are executed
-> return from handle_heredoc to prompt

u413-284-si:
Handles heredoc behaviour when receiving
1. Ctrl + c: Entire pipe is cancelled, no cmd is executed.
2. Ctrl + d: Heredoc receives EOF, an error message is printed, the pipe is executed nonetheless.

To achieve this:
* Opening of outfiles moved from parser to executor. When opened in parser, non-existing outfiles will already be created while heredoc is waiting for input in contrast to bash. On cancellation all outfiles would need to be deleted. If outfiles were already existent in beforehand, those would be deleted as well which is not acceptable, as there is no way to know which were existent and which created. By moving the opening to the cmd execution the outfiles will not be created in the first hand as the heredoc handling happens before the execution and will terminate the execution chain upon cancellation.
* Opening of outfiles in executor requires saving of outfile names as well as saving the information which outfile is opened in append mode. This information is saved in the parser into respectively a str array and bool array.
* Error handling in both cases implemented to achieve desired behaviour.
* Prototype cleanup function for executor implemented.
* Protection of execution process if no outfiles are present.
* File descriptor redirection and resetting to STDIN and STDOUT for singular builtins supplemented.

Bugfix:
* Child processes that did not execute via execve (=> built-ins) are now terminated with exit(0) and do not continue running with the parent process

Closes #49 #53


.vscode/launch.json
```diff
@@ -4,6 +4,25 @@
 	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
 	"version": "0.2.0",
 	"configurations": [
+		{
+			"name": "(gdb) Attach",
```
gwolf-011235, line 8: Did attach config work with minishell? I hang in readline. We could add the config which follows the child process

inc/minishell_error.h
```diff
@@ -57,7 +57,8 @@ typedef enum e_error {
 	ERR_OPEN, ///< open() failed.
 	ERR_SIGNAL, ///< Setup of signal handler failed.
 	ERR_NO_INPUT, ///< Input is empty.
-	ERR_ABORT ///< Abort pipe.
+	ERR_ABORT, ///< Abort pipe.
```
gwolf-011235: I'd clarify "Abort pipeline" since "Abort pipe" is more akin to singular pipe

inc/mod_handle_input.h
```diff
@@ -34,25 +34,40 @@
...
+	int				out_pos;
+	bool			*append_switches;
 	int				fd_in;
 	int				fd_out;
 	int				fd_pipe[2];
 	int				fd_prev_pipe[2];
-	char			**delims;
-	int				delim_pos;
-	char			*heredoc;
 	bool			append;
```
gwolf-011235: I think since there is now an array of bools saving which outfile to open in append mode, this singular bool can be eliminated. It was for testing purposes, but the test should be adapted to work with array I'd say

src/executor.c
```diff
@@ -107,8 +117,20 @@ t_err	ft_execute_pcmds(t_cmd *cmd,
 {
 	t_err	err;
 	t_cmd	*tmp;
+	t_cmd	*tmp2;
 
 	tmp = cmd;
+	tmp2 = cmd;
```
gwolf-011235: You can eliminate tmp2, if you just reassign tmp to cmd. 

Maybe rename tmp to cmd_head, so it's clear that we are at start of list. Then you could save cmd in cmd_head - loop through cmd for outfiles - reassing cmd to cmd_head - loop through for exe - wait for cmd_head.

src/executor_builtin.c
```diff
@@ -49,15 +52,64 @@ bool	ft_check_builtin(char *arg)
 t_err	ft_execute_builtin(bool piped, t_cmd *cmd, t_data *data)
 {
 	t_err	err;
+	int		old_stdin;
+	int		old_stdout;
 
+	old_stdin = 0;
+	old_stdout = 0;
```
gwolf-011235: Why set to 0 here and not dup(STDIN_FILENO) and dup(STDOUT_FILENO).

If reason, then you'd have to pass these variables as pointers to ft_set_fd_scmd(), else nothing happens.

src/executor_builtin.c
```diff
@@ -49,15 +52,64 @@ bool	ft_check_builtin(char *arg)
...
+ * 
+ * @param cmd 		Current cmd.
+ * @return t_err 	ERR_CLOSE, SUCCESS
+ */
+t_err	ft_set_fd_scmd(t_cmd *cmd, int old_stdin, int old_stdout)
+{
+	t_err	err;
+
+	old_stdin = dup(STDIN_FILENO);
+	old_stdout = dup (STDOUT_FILENO);
```
gwolf-011235: See comment above. This is only local for the function

src/executor_builtin.c
```diff
@@ -49,15 +52,64 @@ bool	ft_check_builtin(char *arg)
...
+/**
+ * @brief Reset STDIN_FILENO and STDOUT_FILENO.
+ * 
+ * Stdin and Stdout have been replaced by other fds.
+ * They are reset to STDIN_FILENO and STDOUT_FILENO with backup.
+ * @return t_err 	ERR_CLOSE, SUCCESS
+ */
+t_err	ft_reset_fd_scmd(int old_stdin, int old_stdout)
+{
+	ft_replace_fd(old_stdin, old_stdout);
```
gwolf-011235, line 114: Doxy says ERR_CLOSE but it to only returns SUCCESS. I think ft_replace_fd() needs to be catched?

src/executor_child.c
```diff
@@ -27,35 +27,38 @@
...
+		ft_close(&cmd->fd_pipe[1]);
 	}
-	ft_close(cmd->fd_in);
+	ft_close(&cmd->fd_in);
 	if (builtin)
 		ft_choose_builtin(cmd, data);
 	else
 		if (execve(cmd->args[0], cmd->args, data->envp) < 0)
 			printf("\nexecve encountered an error\n");
+	exit(0);
```
gwolf-011235, line 61: Should return g_status, so we can use exit status of builtins. 
Applies to all three functions

Builtins need to be adapted, so they set g_status. I will open an issue for this

src/executor_child.c
```diff
@@ -27,35 +27,38 @@
...
-		ft_close(cmd->fd_pipe[1]);
+		ft_close(&cmd->fd_pipe[1]);
 	}
-	ft_close(cmd->fd_in);
+	ft_close(&cmd->fd_in);
 	if (builtin)
 		ft_choose_builtin(cmd, data);
 	else
 		if (execve(cmd->args[0], cmd->args, data->envp) < 0)
 			printf("\nexecve encountered an error\n");
```
gwolf-011235, line 60: Maybe mark this printf call, so that we don't forget to replace it with an error message printed to stderr

src/executor_cleanup2.c
```diff
@@ -0,0 +1,36 @@
...
+ */
+void	ft_cleanup_cmd_list(t_cmd *cmd)
+{
+	t_err	err;
+
+	while (cmd)
+	{
+		err = ft_cleanup_cmd(cmd);
+		if (err != SUCCESS)
+			perror("Closing of fd failed.");
```
gwolf-011235, line 33: I think this would be better in it's own syscall wrapper functions as discussed. In this case: ft_cleanup_cmd calls unlink, which can fail for a number of errno reasons.

src/executor_create_child.c
```diff
@@ -65,7 +65,8 @@ t_err	ft_raise_first(t_cmd *cmd, t_data *data, bool builtin)
 		ft_firstborn(cmd, data, builtin);
 	if (signal(SIGINT, SIG_IGN) == SIG_ERR)
```
gwolf-011235: Here we can use our ft_signal_setup. 
Since we agreed to not do anything if signal assignage does not work, we don't need error check.

Applies to other functions as well

src/executor_heredoc.c
```diff
@@ -68,7 +68,7 @@ t_err	ft_create_heredoc(t_cmd *cmd, char *delim, int curr_delim,
 		return (err);
 	g_status = 0;
 	err = ft_read_heredoc(delim, prompt2, fd, &name);
-	if (err != SUCCESS)
+	if (err != SUCCESS && err != ERR_HEREDOC_EOF)
```
gwolf-011235, line 71: What do you think of writing it positive: if (err == ERR_ABORT)?

src/executor_heredoc.c
```diff
@@ -99,7 +99,7 @@ t_err	ft_read_heredoc(char *delim, char *prompt2, int fd, char **name)
 		if (g_status == 130)
 			return (ft_unlink_heredoc(name, ERR_ABORT));
 		if (!buf)
-			return (ft_print_warning(delim, prompt2));
+			return (ft_print_warning(delim));
 		if (ft_strncmp(delim, buf, len + 1) == 0)
 			break ;
 		write(fd, buf, ft_strlen(buf));
```
gwolf-011235, line 108: Since readline eats the newline \n at the end of input, we need to rewrite it after this write()

src/executor_heredoc2.c
```diff
@@ -47,16 +47,13 @@ t_err	ft_initiate_heredoc(int index, char **name, int *fd)
  * Then print the warning message with
  * expected the delimiter.
  * @param delim		Delimiter string of heredoc.
- * @param prompt2	Heredoc prompt.
- * @return t_err	SUCCESS
+ * @return t_err	ERR_HEREDOC_EOF
  */
-t_err	ft_print_warning(char *delim, char *prompt2)
+t_err	ft_print_warning(char *delim)
```
gwolf-011235, line 52: I think print_warning is very generic. Maybe we can incorporate this in a general warning function. If not we should rename this.

src/executor_utils2.c
```diff
@@ -108,3 +108,34 @@ bool	ft_check_empty_path(char *path_str)
...
+		if (cmd->fd_out > -1)
+			if (close(cmd->fd_out) < 0)
+				return (ERR_CLOSE);
+		if (cmd->append_switches[i])
+			cmd->fd_out = open(cmd->outfiles[i], O_WRONLY | O_APPEND | O_CREAT,
+					0644);
+		else
+			cmd->fd_out = open(cmd->outfiles[i], O_RDWR | O_TRUNC | O_CREAT,
+					0644);
+		if (cmd->fd_out == -1)
```
gwolf-011235, line 137: I think it would make sense to always check for == -1, if it's an syscall error check. Right now there is also - if not more often - smaller zero (e.g. right here line 129). We could differentate between syscall errors (== -1) and not assigned cmd->fd ( > -1)

I'd not hunt the code for it, just if an instance is caught, make it uniform.

src/handle_input.c
```diff
@@ -54,6 +54,6 @@ t_err	ft_handle_input(char *input, t_data *data)
 	ft_free_lst(&lst);
 	err = ft_executor(cmd, data);
 	if (err != SUCCESS)
-		return (err);
+		ft_cleanup_cmd_list(cmd);
```
gwolf-011235, line 57: I think cleanup should run always regardless of success? 
Maybe ft_cleanup_cmd_list could bubble the error of ft_executor up (receives and returns it).

src/parser_fill_cmd.c
```diff
@@ -78,33 +78,33 @@ t_err	ft_save_heredoc(t_tkn_list **lst, t_cmd *new)
  * @brief Open file and save fd as fd_out.
  *
  * If append bool is true, open file in append mode.
+ * Names of outfiles are saved in string array.
  * @param lst 		List of token.
  * @param new 		New cmd to be filled.
  * @param append	Bool to indicate if append mode is required.
  * @return t_err 	ERR_OPEN, SUCCESS
```
gwolf-011235: Needs to be changed, since this doesn't open anymore.

src/parser_fill_cmd.c
```diff
@@ -78,33 +78,33 @@ t_err	ft_save_heredoc(t_tkn_list **lst, t_cmd *new)
...
  */
 t_err	ft_save_outfile(t_tkn_list **lst, t_cmd *new, bool append)
 {
-	int			fd_out;
 	t_tkn_list	*tmp;
 
 	tmp = *lst;
```
gwolf-011235: Why do we need tmp, if we move the list forward anyways?

src/parser_fill_cmd.c
```diff
@@ -78,33 +78,33 @@ t_err	ft_save_heredoc(t_tkn_list **lst, t_cmd *new)
...
-	if (fd_out == -1)
-		return (ERR_OPEN);
-	new->fd_out = fd_out;
+	new->outfiles[new->out_pos] = ft_strdup(tmp->content);
+	if (!new->outfiles[new->out_pos])
+		return (ERR_MALLOC);
 	if (append)
+	{
+		new->append_switches[new->out_pos] = true;
 		new->append = 1;
```
gwolf-011235: I think this switches on and off everytime append is (not) found

So append, append, open would result in new->append = 0

src/parser_utils.c
```diff
@@ -102,6 +109,34 @@ t_err	ft_create_str_arr(t_cmd *tmp, int count_arg, int count_delim)
...
+ */
+t_err	ft_create_str_arr2(t_cmd *tmp, int count_out)
+{
+	if (count_out)
+	{
+		tmp->outfiles = malloc(sizeof(char *) * (count_out + 1));
+		if (!tmp->outfiles)
+			return (ERR_MALLOC);
+		tmp->outfiles[count_out] = NULL;
+		tmp->append_switches = malloc(sizeof(bool) * count_out);
```
gwolf-011235: Needs to free tmp->outfiles if fails

src/parser_utils.c
```diff
@@ -102,6 +109,34 @@ t_err	ft_create_str_arr(t_cmd *tmp, int count_arg, int count_delim)
...
+}
+
+/**
+ * @brief Reserve memory space for outfiles string array.
+ * 
+ * @param tmp 		New cmd struct to be created.
+ * @param count_out Amount of outfiles.
+ * @return t_err 	ERR_MALLOC, SUCCESS
+ */
+t_err	ft_create_str_arr2(t_cmd *tmp, int count_out)
```
gwolf-011235: ft_create_str_arr_outfile(). Just to be a little bit more verbo~se

src/parser_utils.c
```diff
@@ -102,6 +109,34 @@ t_err	ft_create_str_arr(t_cmd *tmp, int count_arg, int count_delim)
...
+		tmp->outfiles[count_out] = NULL;
+		tmp->append_switches = malloc(sizeof(bool) * count_out);
+		if (!tmp->append_switches)
+			return (ERR_MALLOC);
+	}
+	else
+	{
+		tmp->outfiles = NULL;
+		tmp->append_switches = NULL;
+	}
```
gwolf-011235: This (and the other inits) could be stuffed into ft_init_cmd(). Would save couple o'lines

src/executor_cleanup.c
```diff
@@ -30,92 +30,26 @@ t_err	ft_cleanup_cmd(t_cmd *cmd)
...
 	if (err != SUCCESS)
 		return (err);
 	if (cmd->heredoc)
 		ft_unlink_heredoc(&cmd->heredoc, err);
+	if (cmd->delims)
+		ft_free_str_arr(cmd->delims);
+	if (cmd->args)
+		ft_free_str_arr(cmd->args);
+	if (cmd->outfiles)
+		ft_free_str_arr(cmd->outfiles);
```
gwolf-011235, line 49: Outfile bool array also needs to be free'd

gwolf-011235 (changes requested):
Nice job of re-integrating the already implemented outfile opening - looks good as far as I can tell!
Lets hope that the builtin shenanigans are coming to an end now.

There are some potential leaks.
I also think that cleanup_cmds needs to be called in any case not just error.

src/parser_fill_cmd.c
gwolf-011235: Which - after some sleep - makes sense, since there can only be one outfile :D But a heredoc in the chain would break it I think
outfile, append, heredoc = append is true

.vscode/launch.json
u413-284-si: Requires sudo rights :(

inc/minishell_error.h
u413-284-si: Done

inc/mod_handle_input.h
u413-284-si: bool append deleted

.vscode/launch.json
gwolf-011235: Oh no :(

src/executor.c
u413-284-si: removed tmp2

src/executor_builtin.c
u413-284-si: Setting moved to parent function

src/executor_builtin.c
u413-284-si: see above

src/executor_builtin.c
u413-284-si: Included closing of oldstdin/oldstdout. After setting them in parent function their closing doesn't lead to program exit anymore

src/executor_child.c
u413-284-si: Will be addressed in extra branch

src/executor_child.c
u413-284-si: done

src/executor_cleanup2.c
u413-284-si: only temporary prototype, needs to be elaborated in cleanup branch

src/executor_create_child.c
u413-284-si: Replaced

src/executor_heredoc.c
u413-284-si: Will have more ERR returns with heredoc expander which require return as well

src/executor_heredoc.c
u413-284-si: si senor

src/executor_heredoc2.c
u413-284-si: Will be specified in error handling branch

src/executor_utils2.c
u413-284-si: :)

src/handle_input.c
u413-284-si: will be elaborated later in cleanup branch

src/parser_fill_cmd.c
u413-284-si: updated

src/parser_fill_cmd.c
u413-284-si: Using dereferenced lst now

src/parser_fill_cmd.c
u413-284-si: sleep no good

u413-284-si: heredoc => infile
append => outfile

src/parser_utils.c
u413-284-si: General cleanup will free all malloced variables

src/parser_fill_cmd.c
gwolf-011235: true

src/parser_utils.c
u413-284-si: Created extra malloc function

src/parser_utils.c
u413-284-si: Switched to init function

src/executor_cleanup.c
u413-284-si: supplemented

gwolf-011235 (approved):
Nice - ready to merge! :rocket:

PR #60, opened 2023-08-15
feat: check for directories

### LINKED ISSUE #52: Recognise dir
Need to recognise if given cmd is a dir inside the current directory. Print out an appropriate err message.

Example:
bash-5.1$ core_okt_2022/
bash: core_okt_2022/: Is a directory

Can be handled with stat() function.

u413-284-si:
Added ability to check given input (args) for directories utilising stat() and its Macro S_ISDIR:

* If contains '/': check for directory
If yes, print according message "Is a directory". Exit status = 126
If not found, print according message "No such file or directory". Exit status 127
* Else continue to check if it's a vaild cmd. 
If yes, execute cmd. 
If not, print according message "command not found". Exit status 127

Has potential to refactor by pulling the directory check out of ft_check_cmd_access(). Follows after refactor of ft_execute_scmd() and ft_execute_pcmd()

Close #52

### DIFF
--- a/inc/minishell_error.h
+++ b/inc/minishell_error.h
@@ -6,7 +6,7 @@
 /*   By: sqiu <sqiu@student.42vienna.com>           +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/05/24 11:58:28 by gwolf             #+#    #+#             */
-/*   Updated: 2023/08/15 11:34:41 by sqiu             ###   ########.fr       */
+/*   Updated: 2023/08/15 17:02:31 by sqiu             ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
@@ -58,8 +58,12 @@ typedef enum e_error {
 	ERR_SIGNAL, ///< Setup of signal handler failed.
 	ERR_NO_INPUT, ///< Input is empty.
 	ERR_ABORT, ///< Abort pipeline.
-	ERR_HEREDOC_EOF ///< Heredoc stopped by Ctrl+D
+	ERR_HEREDOC_EOF, ///< Heredoc stopped by Ctrl+D
+	ERR_DIR, ///< Cmd is a directory.
+	ERR_NO_DIR ///< Directory of file not found
 }	t_err;
 
+t_err	ft_print_warning(char *indic, char *trigger);
+t_err	ft_print_warning2(char *indic, char *trigger);
 
 #endif
--- a/src/error_temp.c
+++ b/src/error_temp.c
@@ -1,0 +1,60 @@
+
+#include "mod_executor.h"
+
+/**
+ * @brief Print warning messages depending on 
+ * received indicator and trigger string.
+ * 
+ * @param indic		Indicator string for error message.
+ * @param trigger	Trigger string causing the error.
+ * @return t_err	ERR_HEREDOC_EOF, ERR_DIR, SUCCESS
+ */
+t_err	ft_print_warning(char *indic, char *trigger)
+{
+	if (!ft_strncmp(indic, "heredoc", 8))
+	{
+		ft_putstr_fd("minishell: warning: here-document at line 42 \
+	delimited by end-of-file (wanted `", 2);
+		ft_putstr_fd(trigger, 2);
+		ft_putendl_fd("')", 2);
+		return (ERR_HEREDOC_EOF);
+	}
+	else if (!ft_strncmp(indic, "dir", 4))
+	{
+		g_status = 126;
+		ft_putstr_fd("minishell: ", 2);
+		ft_putstr_fd(trigger, 2);
+		ft_putendl_fd(": Is a directory", 2);
+		return (ERR_DIR);
+	}
+	else
+		return (ft_print_warning2(indic, trigger));
+}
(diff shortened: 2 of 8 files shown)


inc/minishell_error.h
```diff
@@ -58,8 +58,12 @@ typedef enum e_error {
 	ERR_SIGNAL, ///< Setup of signal handler failed.
 	ERR_NO_INPUT, ///< Input is empty.
 	ERR_ABORT, ///< Abort pipeline.
-	ERR_HEREDOC_EOF ///< Heredoc stopped by Ctrl+D
+	ERR_HEREDOC_EOF, ///< Heredoc stopped by Ctrl+D
+	ERR_DIR, ///< Cmd is a directory.
+	ERR_NO_DIR ///< Directory of file not found
```
gwolf-011235, line 63: Did you mean Directory OR file not found?

src/error_temp.c
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @brief Print warning messages depending on 
+ * received indicator and trigger string.
+ * 
+ * @param indic		Indicator string for error message.
+ * @param trigger	Trigger string causing the error.
+ * @return t_err	ERR_HEREDOC_EOF, ERR_DIR, SUCCESS
+ */
+t_err	ft_print_warning(char *indic, char *trigger)
```
gwolf-011235, line 12: Why not just use t_err to differentiate the errors? Would get rid of the strncmp() calls.

(If err == ERR_HEREDOC_EOF)
print eof message

src/error_temp.c
```diff
@@ -0,0 +1,60 @@
...
+		ft_putendl_fd(": No such file or directory", 2);
+		return (ERR_DIR);
+	}
+	if (!ft_strncmp(indic, "nocmd", 6))
+	{
+		g_status = 127;
+		ft_putstr_fd("minishell: ", 2);
+		ft_putstr_fd(trigger, 2);
+		ft_putendl_fd(": command not found", 2);
+		return (SUCCESS);
```
gwolf-011235, line 57: Why SUCCESS and not ERR_UNKNOWN_CMD

src/executor_utils.c
```diff
@@ -39,6 +41,10 @@ t_err	ft_check_cmd_access(char **args, char **cmd_paths, bool empty_path)
 {
 	t_err	err;
 
+	if (ft_check_dir(args) == ERR_DIR)
```
gwolf-011235, line 44: If catch return with err you don't have to call the function twice.

src/executor.c
```diff
@@ -158,14 +158,36 @@ t_err	ft_process_cmd(t_cmd *cmd, t_err err, t_data *data)
...
 	{
-		write(2, "minishell: ", 11);
-		ft_putstr_fd(cmd->args[0], 2);
-		write(2, ": command not found\n", 20);
+		ft_print_warning("nocmd", cmd->args[0]);
 		err = ft_close(&cmd->fd_pipe[1]);
 	}
 	else if (err == SUCCESS)
 		err = ft_create_child(cmd, data, false);
 	return (err);
```
gwolf-011235, line 170: This returns with SUCCESS, which means wait_for_children also tries to wait for empty commands.

My guess: In the if condition of waitpid also add (cmd->pid != 0 &&)

Also cmd->pid should be init to 0

gwolf-011235 (commented):
Nice, seems pretty straightforward.

Print_error can be improved with utilization of our beloved t_err

inc/minishell_error.h
u413-284-si: yes, corrected

src/error_temp.c
u413-284-si: Cause not very error has a unique ERR_. This is due to the implementation with ft_check_cmd_access() being followed directly by ft_process_cmd() which takes the returned err to determine its actions. 

As indicated this might be refactored later on when the parent functions ft_execute_scmd() and ft_execute_pcmd() are altered for error handling. A cleaner differentiation between ft_check_dir(), ft_check_access would help this regarding.

src/error_temp.c
u413-284-si: see above, basically when cmd is unknown, we require the pipeline to be continued and this is achieved with SUCCESS. May be changed in refactor

src/executor_utils.c
u413-284-si: see above, will be considered later on

src/executor.c
u413-284-si: What do you mean, that ft_print_warning() returns SUCCESS when "nocmd"?

That's why I don't catch the err of it in that case. But I see the err of ft_close will be SUCCESS usually. Solved if temporarily, but should be adapted with the refactor mentioned above.

cmd->pid initialised to 0

src/error_temp.c
gwolf-011235: I think you already refactored a bit: I currently find four instances of print_warning(). Each in direct response to an error.

src/error_temp.c
gwolf-011235: As you write below you intentionally don't catch the error (executor.c:165). So i doesn't matter from a programming aspect but would make sense to be uniform

PR #61, opened 2023-08-17
58 exit status of builtins using g status

### LINKED ISSUE #58: Exit status of builtins using g_status
Builtins were build before the global g_status existed. They need to be adapted to use g_status.

ft_exit() will need to be adapted. It can't use exit as there still needs to be cleanup done. It could set a flag to break out of loop

gwolf-011235:
- Builtins set g_status to indicate success (0) or an error (usually 1).

Sometimes it was just setting the g_status to 0 at the end (if no errors are possible)
Other builtins required a bit of a rewrite.

- Syscall wrappers for `write` and `chdir` in error_temp.c 

Sets errno to 0, calls function. If return -1 perror the error

- Bool **loop** in main

The input+execute while loop of main depends on bool **loop**. This can be set to false to exit loop and perform cleanup. If a builtin was forked, the bool gets set to false in executor and the child exits and performs cleanup.

- Rewrite of `ft_exit`

`ft_exit` is now based on the loop var in main: `ft_exit` sets it to false. When returning to main, this would exit the loop and perform general cleanup. This means that both builtin and forked `ft_exit` clean on their way out. `ft_exit` only prints exit\n on stderr when acting as a built-in.

- Cleanup of debugger configs

Removed config for ft_split and renamed (gdb) Attach to attach minishell
Four configs are now in launch.json: minishell (debug minishell), test (debug tester), follow child (debug minishell but follow the first fork it creates) and attach minishell (debug an already running minishell, req sudo rights)


Makefile
```diff
@@ -3,10 +3,10 @@
 #                                                         :::      ::::::::    #
 #    Makefile                                           :+:      :+:    :+:    #
 #                                                     +:+ +:+         +:+      #
-#    By: sqiu <sqiu@student.42vienna.com>           +#+  +:+       +#+         #
+#    By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+         #
```
u413-284-si, line 6: yuuuuuuuuuuuuuuuuuuuuuuuuuuuuu

src/builtin_env.c
```diff
@@ -46,5 +46,6 @@ t_err	ft_env(t_hashtable *env_tab)
 		}
 		i++;
 	}
+	g_status = 0;
```
u413-284-si, line 49: nice

u413-284-si (commented):
Nothing to be said, splendid work

PR #63, opened 2023-08-17
62 segault fixes

### LINKED ISSUE #9: Readline
Implement readline function and do stuff with it (add history)

### LINKED ISSUE #62: Segault fixes
There are some segfaults which need to be addressed

gwolf-011235:
Fix two segfaults:

1. `ft_executor`
Bug: if no args no outfiles would be generated ; in pcmd: if no args in cmd after the first one it would try to dereference args[0] which result in segfault
Fix: move the check for args into `ft_execute_scmd` and `ft_execute_pcmd`. This way outfiles are generated everytime. Also if a cmd in pipeline has no args, it just gets skipped for the next cmd
ToDo: does this work the pipes? Or do we need to recheck something here?

2. `ft_check_syntax` and `ft_lexer`
Bug: If input string consists of only spaces, `ft_check_syntax` would let it through. `ft_lexer` would consume all spaces, creating a lst-node which has no string. `ft_assign_type` would try to dereference the string resulting in segfault.
Fix: already check in main if input is empty (= consists of only whitespace). If empty do nothing, and ask again for input. Else `handle_input`. 
This also closes #9 since only meaningful input will be added to history.

Not a segfault but still fix: 

3. `ft_check_dir`
Bug: every absolute path to a cmd (e.g. /bin/ls) would be checked if it is a directory. If not it would return ERR_NO_DIR, stopping execution
Fix: Syscall wrapper for stat `ft_err_stat` which can return SUCCESS (worked), ERR_NO_DIR (errno was ENOENT, which means entity was not found), ERR_STAT (stat failed). 
If not SUCCESS return with err (either stat failed or thing was not found):
Check if directory, if yes return ERR_DIR, else return SUCCESS (which means we found command)


inc/minishell_utils.h
```diff
@@ -24,5 +25,7 @@ void	*ft_realloc(void *ptr, size_t new_size, size_t curr_size);
 //string
 char	*ft_strcat(char *dst, const char *src);
 char	*ft_strncat(char *dst, const char *src, size_t n);
+bool	ft_isspace(char c);
+bool	ft_isempty_str(const char *s);
 
 #endif
```
u413-284-si: our beloved newline wants attention

inc/mod_syntax.h
```diff
@@ -26,6 +27,5 @@ t_err	ft_check_syntax(const char *input);
 t_err	ft_check_redirect(const char *input, size_t pos, char symbol);
 t_err	ft_check_pipe(const char *input, size_t pos);
 t_err	ft_quote_skipper(const char *quote_start, size_t *i, char target);
-int		ft_is_space(char c);
 
 #endif
```
u413-284-si: here as well

src/error_temp.c
```diff
@@ -33,7 +32,7 @@ t_err	ft_print_warning(char *indic, char *trigger)
 
 /**
  * @brief Continuation of ft_print_warning()
- * 
+ *
  * @param indic 	Indicator string for error message.
  * @param trigger 	Trigger string causing the error.
  * @return t_err 	ERR_DIR, SUCCESS
```
u413-284-si: return value is ERR_NO_DIR

src/executor_utils.c
```diff
@@ -41,11 +41,16 @@ t_err	ft_check_cmd_access(char **args, char **cmd_paths, bool empty_path)
 {
 	t_err	err;
 
-	if (ft_check_dir(args) == ERR_DIR)
-		return (ft_print_warning("dir", args[0]));
-	else if (ft_check_dir(args) == ERR_NO_DIR)
-		return (ft_print_warning("nodir", args[0]));
-	err = ERR_UNKNOWN_CMD;
+	if (ft_strchr(*args, '/'))
```
u413-284-si: Would write args[0] for consistency

src/utils_string.c
```diff
@@ -71,3 +71,33 @@ char	*ft_strncat(char *dst, const char *src, size_t n)
...
+ * In the "C" and "POSIX" locales, these are: space, form-feed ('\f'),
+ * newline ('\n'), carriage return ('\r'), horizontal tab ('\t'), and
+ * vertical tab ('\v').
+ * @param c Char to check
+ * @return true Is white-space
+ * @return false Not white-space
+ */
+bool	ft_isspace(char c)
+{
+	if (c == ' ' || c == '\f' || c == '\n' || c == '\r'
```
u413-284-si: Use ft_strchr() maybe?

u413-284-si (commented):
Looks good, small changes and corrections, nothing major.

Moving the check into ft_scmd()  and ft_pcdm() respectively should not break the pipe. If one cmd is not executed, the following cmd should close the connecting pipe and not receive any input. Should be verified by own tests/tester.

inc/minishell_utils.h
gwolf-011235: added

inc/mod_syntax.h
gwolf-011235: added

src/error_temp.c
gwolf-011235: fixed

src/executor_utils.c
gwolf-011235: changed all to args[0] in this ft

src/utils_string.c
gwolf-011235: changed, but also added (== '\0') sanity check for myself :D

u413-284-si: yes my dear

PR #64, opened 2023-08-17
22 last exit status

### LINKED ISSUE #22: Last exit status
Echo $? returns last exit status. 

For example the return code in bash for

- syntax error is 2
- permission denied is 126
- command not found is 127

I think the t_error enum can be used for this. Right now ERR_SYNTAX is set to 2.

gwolf-011235:
Exit status are should be correct now, since g_status is used nearly everywhere, where needed.

- Builtins use g_status to set it to 0, 1, 2 (see #61 )

- 126 and 127 are handled by `ft_print_warning`

- This PR changes behaviour of `ft_syntax_check` to use g_status

- Not implemented: bash truncates the number passed to builtin `exit` to 8 bits. Bash doesn't check for int (32 bit) overflow, but for long (64 bit) overflow and writes: numeric argument required. POSIX leaves the behaviour undefined, if the value passed to special builtin exit is outside of [0 - 255]. We just use libft `ft_atoi` on the number, which can return a greater number than 255 (which on turn would be then truncated by bash, if called from bash). On int overflow we return -1 or 0. Can be adapted if time is there but since not defined, I'd leave it as is.

This closes #22


inc/mod_syntax.h
```diff
@@ -6,26 +6,41 @@
...
 
+//syntax.c
 t_err	ft_check_syntax(const char *input);
 t_err	ft_check_redirect(const char *input, size_t pos, char symbol);
 t_err	ft_check_pipe(const char *input, size_t pos);
 t_err	ft_quote_skipper(const char *quote_start, size_t *i, char target);
 int		ft_is_space(char c);
+t_err	ft_syntax_error(t_syntax syntax_err, char c);
 
 #endif
```
u413-284-si: neewline

u413-284-si (commented):
Quite straightforward. Nothing to be added.

No need to implement the truncation feature in exit(). Due to time constraints we should focus on explicit requirements and leave anything unspecified aside.

inc/mod_syntax.h
gwolf-011235: fixed in #62

u413-284-si:
ulaaa, some merge conflicts arised from prev merge. Once resolved PR may be merged by assignee as no open points remain

PR #65, opened 2023-08-17
27 make buffer extra

### LINKED ISSUE #27: Make buffer extra
Right now the buffer in the tokenising step gets malloc'd for every token created. I think it would be crucial to malloc it once and then use it again and again, until every token is created.

I think this would increase the tokenisation speed by about 100% 👯‍♂️

gwolf-011235:
Data struct now has a t_buf buf.

- At startup, the buffer gets malloced at the specified size. This buffer can be passed around. At exit (outside of the loop) the buffer gets destroyed.
- Right now the functions `ft_lex_input` and `ft_expand_tkn_list` get the buffer from data. They don't need to malloc it for every call, can reuse it. If the buffer size is too small, these functions have the ability to double the buffer size with `ft_buf_double`. This doubling will persist until program shutdown.
- `ft_add_to_buf` doesn't need to realloc anymore if buffer is full. This makes malloc error handling a little bit easier.

Litte touchup on the current layout of main. 
- Create buffer at start, destroy it at end
- Create envp at start of every loop, destroy it at end of loop. 

Closes #27


src/lexer_tok_utils.c
```diff
@@ -32,32 +32,22 @@ void	ft_free_tok(t_tok *token)
 
 /**
  * @brief Add character to buffer string.
- * 
+ *
  * In case the buffer string is not long enough, its size is doubled.
  * @param c 		Character to be added.
  * @param buf 		Temporary buffer to save as token.
- * @return t_err 	ERR_MALLOC, SUCCESS
+ * @return t_err 	SUCCESS
```
u413-284-si: Function doesn't need to return anything anymore as return is not caught in ft_partition()

u413-284-si:
Branch has some conflicts which need resolving before merge can be put into effect

src/lexer_tok_utils.c
gwolf-011235: changed to void

u413-284-si:
here as well, given all conflicts are resolved, PR can be merged by assignee himself

PR #67, opened 2023-08-19
17 implement special parameters

### LINKED ISSUE #17: Implement special parameters
Bash has special parameters to which assignment is not allowed (see 3.4.2 Special Parameters)

Some seem to be easily implemented, others not that much. Doable should be the following:

- [x] ? - Expands to the exit status of the most recently executed foreground pipeline.
- [ ] $ - Expands to the process ID of the shell. In a subshell, it expands to the process ID of the invoking shell, not the subshell.
- [ ] 0 - Expands to the name of the shell or shell script. This is set at shell initialization. If Bash is invoked with a file of commands (see [Shell Scripts](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Scripts)), $0 is set to the name of that file.

gwolf-011235:
Special parameter `$$` and `$0` have been implemented.

-  New member `export` in env variable struct
Env variables now track export status. Nearly every variable gets this status except special parameters $$ and $0.
New hashtable functions to set this status. Also a wrapper to insert var and set export with one function.

- Special params are inserted at startup
Get pid and insert variable $$
Get argv[0] and insert variable $0

- `ft_expand_dollar` adapted
Expander now recognises special variables $$ and $0 and expands them.

- PS1 and PS2 can now be inserted at startup
The two prompt strings are inserted at startup but don't get export status. This makes something neat possible: Since these are "normal" variables, they can be changed afterwards. They can also be unset, after they have been exported. 

**ToDo**: setup function needs to be adapted, to better handle errors. This will be done in another branch


inc/minishell_config.h
```diff
@@ -19,6 +19,8 @@
 # define MAX_SHLVL 10
 # define BUF_SIZE 2048
 
+# define PS1 "PS1=\\u@\\h:\\w$ "
```
u413-284-si, line 22: The first backslash is to escape the variables \u, \h, \w I presume? Why doesn't $ need to be escaped?

src/env_pid.c
```diff
@@ -0,0 +1,90 @@
...
+ * @brief Retrieve pid for current process.
+ *
+ * /proc/self/stat provides status information about the process.
+ * /self/ is a magic link pointing to the calling process.
+ * Open, read with get_next_line().
+ * First number corresponds to pid. Convert with ft_atoi().
+ * @param pid Pointer to int where to save.
+ * @return t_err SUCCESS, ERR_OPEN, ERR_MALLOC
+ */
+t_err	ft_get_pid_value(int *pid)
```
u413-284-si: pid has special unit pid_t. It is a signed int especially created to save pids. Usage would be in this case:

pid_t *pid

src/env_pid.c
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * @brief Creates env_var $$=PID.
+ *
+ * @param env_pid Where to save env_str
+ * @return t_err SUCCESS, ERR_OPEN, ERR_MALLOC
+ */
+t_err	ft_create_env_pid(char **env_pid)
+{
+	t_err	err;
+	int		pid;
```
u413-284-si: see above

src/env_pid.c
```diff
@@ -0,0 +1,90 @@
...
+	if (!(*env_pid))
+		return (ERR_MALLOC);
+	return (SUCCESS);
+}
+
+/**
+ * @brief Creates and inserts the env_var $$.
+ *
+ * @param env_table Environment.
+ * @return SUCCESS, ERR_MALLOC
```
u413-284-si: misses ERR_OPEN

src/env_pwd.c
```diff
@@ -83,3 +83,25 @@ t_err	ft_create_env_pwd(char **pwd)
 		return (ERR_MALLOC);
 	return (SUCCESS);
 }
+
+/**
+ * @brief Creates and inserts the env_var PWD.
+ *
+ * If no PWD is present in env_table, this function inserts.
+ * @param data
+ * @return t_err If SUCCESS it exits
```
u413-284-si: would adapt it to enlist all return t_err types as we do usually

src/env_setup.c
```diff
@@ -49,6 +49,9 @@ t_err	ft_env_setup(t_hashtable **env_table)
 		err = ft_insert_env_shlvl(*env_table);
 	if (err != SUCCESS)
 		return (err);
+	err = ft_insert_env_pid(*env_table);
```
u413-284-si, line 52: err handling in its own branch as mentioned, right?

u413-284-si (commented):
Nice add-on ;) 

The benefit of adding the export designation has escaped my comprehension though. How does this contribute to the implementation of the special parameters?

inc/minishell_config.h
gwolf-011235: We are in another context: prompt creation. Here we don't have shell variables like $USER. We have predefined escape sequences like \u and \w. 

This means u is just the letter u. But escaped with backslash: \u it will be replaced with the variable $USER 
(btw real bash uses getuid() or something and doesn't rely on shell variables - so it's even more removed from shell variables)

$ has no special meaning in this context, not as a letter nor as sequence \$

src/env_setup.c
gwolf-011235: Yep, it'll get moved into its own function

src/env_pid.c
gwolf-011235: Will replace soon

gwolf-011235:
We don't want to export the special parameters into environment, i.e. we don't want to export pid and shell name to the next program.

Therefore we need a way to differentiate between variables to export and not to. 
This was easily achieved with the bool export :D

inc/minishell_config.h
u413-284-si: Ok, but why don't we just right \u@\h:\w$ then?

What is the double back slash for?

gwolf-011235: Cause of C language: 

- While writing a program

a single backslash is used for escaping. For the newline character we write \n. If we want to have a regular backslash we need to escape it and write \\\\.

Just \u would throw an error because the character '\u' does not exist.

- While receiving input from keyboard

When receiving input from keyboard, it will already be escaped - the character received is '\\\\'

If we are in shell and redefine PS1 / PS2 we can write "export PS1=hi \u ". It will be "seen" as "export PS1=hi \\\\u ". 

That's also the reason why it's not possible to write a newline character as just \n: it will be received as \\\\n which means it will be seen as backslash and the letter n

u413-284-si: That's what I meant in my first comment: 

so in the prompt env->

u is just u
\u is $USER
\\u is ...?

src/env_pid.c
gwolf-011235: Replaced

src/env_pid.c
gwolf-011235: Replaced

src/env_pid.c
gwolf-011235: Added

src/env_pwd.c
gwolf-011235: Updated

inc/minishell_config.h
gwolf-011235: It seems that on GitHub the backslash also has a special meaning. To write two backslashes you need to write four: \\\\

I updated the above comment. I hope it's more clear now.

u413-284-si: A bit, but let's talk about it when we meet again.

PR #68, opened 2023-08-19
66 bug pipelines are stuck

### LINKED ISSUE #66: Bug: pipelines are stuck
Some pipelines like `env | sort` and `export TEST=hi | env | grep TEST` are stuck.

The most likely explanation has to do with builtins: right now they continue through the "parent" flow, after executing the builtin. Before the builtins used `exit`, but this would show leaks in valgrind (builtin process needs to do cleanup). 

Three faulty places were identified so far:

- [x] After `fork` the parent flow should be protected in an else clause
- [x] After coming back to the execution loop the builtin has to break out of loop 
- [x] The builtin should not go into `ft_wait_for_babies`

gwolf-011235:
After fork child process safely returns to main and cleans up

- parent flow after fork is protected via else clause
- child breaks after coming back to exec loop
- child breaks out of wait_for_babies before doing anything

### DIFF
--- a/src/executor.c
+++ b/src/executor.c
@@ -6,7 +6,7 @@
 /*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/07/31 11:04:05 by sqiu              #+#    #+#             */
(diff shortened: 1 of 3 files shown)


u413-284-si:
Nice work, the else protection is more clean this way

PR #70, opened 2023-08-25
69 nonexist path for outfile aborts command

### LINKED ISSUE #69: Nonexist path for outfile aborts command
If a redirect into outfile fails, because a subdirectory does not exist, it causes the pipeline to abort.

`echo hi >./noexist/out | echo bye`

Expect:
bye
minishell: ./noexist/out: No such file or directory

Actual:
_nothing_

In general does an error in outfile creation abort the whole pipeline. It should only have an effect on the current command in pipeline.

u413-284-si:
Handles various error cases:

* does not abort pipeline if invalid (path to) outfile is encountered
* closes all pipelins
-> Parent closes step by step the previous pipe of the child just created
-> children individually close all inherited pipes
-> closing of pipes if a cmd is not executed
* parent waits for all children, even when first cmd is not executed
* does not abort pipeline if invalid (path to) infile is encountered

Closes #69


gwolf-011235 (commented):
Looks good - nice work!

PR #75, opened 2023-08-27
73 ambiguous redirect

### LINKED ISSUE #73: Ambiguous redirect
Needs to be handled: print out error msg if ambiguous redirect encountered.

gwolf-011235:
Handle ambiguous redirects:
- Save filename before expanding, print error message when empty expand and set node type to AMBIGUOUS
- Parser ignores AMBIGUOUS node and next one, while counting arg size and sets cmd to no execute


u413-284-si:
Very nice, nothing to be added :)

PR #79, opened 2023-08-27
78 concat vars

### LINKED ISSUE #78: Concat vars
Use += to concatenate variables

gwolf-011235:
Implements concatenating of variable values with string magic!


u413-284-si:
Requires documentation but other than that looks good.

